PR #14, opened 2025-11-21
feat: add token usage logging & aggregator

rushikeshadhav:
- add token_logger.py for safe token/latency logging
- add tools/aggregate_tokens.py to summarize logs to CSV
- instrument agent.py to log planner LLM calls (tokens + latency)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,5 @@ blog_draft*.md
 # claude
 devcontainer.json
 .claude/
+
+examples/analyzer/token_efficient_agent/logs/usage.log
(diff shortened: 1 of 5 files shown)


meAmitPatil:
Hey @rushikeshadhav Thanks for putting this together and thinking about token usage monitoring, that's definitely an important aspect we need to address.

However, we'd prefer not to handle logging this way. We're planning to integrate proper observability tools like telemetry or LangTrace for logging and monitoring.

We'll be putting out some good first issues soon, so if you're interested in contributing, keep an eye out! We'd love to have you involved in the project.

rushikeshadhav:
Sure, I would be happy to contribute.

PR #25, opened 2025-12-04
Amit/onboarding improvements

meAmitPatil:
1) `rayai init` now automatically installs project dependencies (use `--no-install` to skip)
2) Changed "deploy" → "run"
3) Clarify RayAgent Class benefits

### DIFF
--- a/src/ray_agents/cli/commands/init.py
+++ b/src/ray_agents/cli/commands/init.py
@@ -86,7 +86,7 @@ def init(project_name: str, project_type: str, no_install: bool):
                     click.echo("  pip install -r requirements.txt")
                 click.echo("  # Edit .env file with your API keys")
                 click.echo("  # Create your first agent: rayai create-agent <name>")
-                click.echo("  # Deploy agents: rayai serve")
+                click.echo("  # Run agents: rayai serve")
 
             except Exception as e:
                 click.echo(f"Error: Failed to create project: {e}")
--- a/src/ray_agents/cli/commands/init.py
+++ b/src/ray_agents/cli/commands/init.py
@@ -1,6 +1,8 @@
 """Initialize new projects with templates."""
 
 import shutil
+import subprocess
+import sys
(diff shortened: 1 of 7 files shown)


src/ray_agents/cli/commands/init.py
```diff
@@ -16,7 +18,12 @@
     type=click.Choice(["agent"]),
     help="Type of project to initialize (default: agent)",
 )
-def init(project_name: str, project_type: str):
+@click.option(
+    "--no-install",
```
pavitrabhalla: In what cases would someone want to skip installing dependencies? Is it just `ray` dependencies that are installed? If there's no useful cases for skipping dependencies, I would just install them by default on `init` / minimize the steps to value.

pavitrabhalla (commented):
Left one comment

src/ray_agents/cli/commands/init.py
meAmitPatil: Yes, I agree we are just installing agentic-ray during the init. There is no need to skip install.

src/ray_agents/cli/templates/agent/README.md
```diff
@@ -158,6 +155,8 @@ class MyAgent(RayAgent):
 
 ## Resource Configuration
 
+Specify CPU, GPU, and memory requirements directly in your Python code—no Dockerfiles or Kubernetes manifests needed.
+
```
pavitrabhalla, line 159: Are all these resource configuration instructions still true?


```suggestion
```
This may not be true as we add Kubernetes ^

pavitrabhalla (approved):
Added a comment unrelated to changes in this PR

src/ray_agents/cli/templates/agent/README.md
meAmitPatil: Yes I will make a separate PR, to remove all the resource configs from the codebase and use defaults.

PR #34, opened 2025-12-11
test: add integration tests for tool adapters and parallel execution

meAmitPatil:


tests/test_langchain_with_langchain_tools.py
```diff
@@ -0,0 +1,102 @@
...
+        def my_tool(x: str) -> str:
+            """My tool docstring."""
+            return x
+
+        ray_tool = from_langchain_tool(my_tool)
+
+        adapter = ToolAdapter(framework=AgentFramework.LANGCHAIN)
+        wrapped_tools = adapter.wrap_tools([ray_tool])
+
+        assert wrapped_tools[0].__name__ == "my_tool"
```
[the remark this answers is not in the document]
pavitrabhalla: @meAmitPatil seems like a good catch. What do you think?

tests/test_parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+
+        results = ray.get([remote_1.remote(5), remote_2.remote(5)])
+
+        assert results[0]["result"] == 10  # 5 * 2
+        assert results[1]["result"] == 15  # 5 * 3
+
+        # Both tools should start within 0.15s of each other (parallel)
+        # If sequential, second would start ~0.3s after first
+        time_diff = abs(results[0]["start_time"] - results[1]["start_time"])
+        assert (
```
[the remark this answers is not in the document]
pavitrabhalla: this is also a reasonable concern. there should be a better way to test parallel execution

tests/test_parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+        ]
+
+        # Parallel execution
+        parallel_results = execute_tools(tool_calls, parallel=True)
+
+        assert parallel_results[0]["result"] == 2
+        assert parallel_results[1]["result"] == 4
+        assert parallel_results[2]["result"] == 6
+
+        # All 3 tools should start within 0.15s of each other (parallel)
```
pavitrabhalla, line 118: yeah these time based tests are going to be brittle

tests/test_parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+                else:
+                    msg = AIMessage(content="Done! Results: 10 and 15")
+                return ChatResult(generations=[ChatGeneration(message=msg)])
+
+        mock_llm = MockChatModel()
+        agent = create_react_agent(mock_llm, lc_tools)
+
+        # Run the agent
+        await agent.ainvoke(
+            {"messages": [HumanMessage(content="Run both tools with x=5")]}
```
pavitrabhalla, line 232: without a batch tool, the LLM is always going to call tools sequentially. They could still be within the bounds of time difference you're testing for. I think this test is misleading.

tests/test_pydantic_with_langchain_tools.py
```diff
@@ -0,0 +1,110 @@
...
+        # Convert to Ray tool
+        ray_weather = from_langchain_tool(get_weather)
+
+        # Wrap with ToolAdapter for Pydantic
+        adapter = ToolAdapter(framework=AgentFramework.PYDANTIC)
+        wrapped_tools = adapter.wrap_tools([ray_weather])
+
+        assert len(wrapped_tools) == 1
+
+        result = wrapped_tools[0](city="San Francisco")
```
pavitrabhalla, line 33: how do we know that calling wrapped_tools this way is mimicing how the Pydantic agent will call tools? I think this is right but its unclear if it really is.

PR #46, opened 2025-12-13
Amit/batch feature

meAmitPatil:
## Testing

- [ ] Tested locally
- [ ] Added/updated tests
- [ ] All tests passing

### DIFF
--- a/src/ray_agents/batch.py
+++ b/src/ray_agents/batch.py
@@ -12,7 +12,6 @@ from ray_agents.adapters.core import RayTool, to_raytool
 
 def batch(
     tool: Callable | RayTool,
-    max_concurrency: int | None = None,
     description: str | None = None,
 ) -> Callable:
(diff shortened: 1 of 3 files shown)


src/ray_agents/batch.py
```diff
@@ -0,0 +1,225 @@
...
+    Returns:
+        A new callable that accepts list inputs and executes in parallel.
+        Can be wrapped with RayToolWrapper for framework-specific format.
+
+    Example:
+        >>> from ray_agents import tool, batch
+        >>>
+        >>> @tool(desc="Fetch content from a URL")
+        >>> def fetch_url(url: str) -> str:
+        ...     return requests.get(url).text
```
pavitrabhalla: How does passing inputs work when there are multiple arguments to the tool? It may be useful to showcase or document such an example

PR #164, opened 2026-04-16
Docs & SDKs: Fern-generated  → hand-crafted

nirnejak:
## Description

Replaces the Fern-generated TypeScript and Python SDKs with hand-crafted implementations centered on a `Sandbox` class API. Removes all Fern infrastructure (~15k lines of generated code) and replaces it with minimal, focused modules: `Sandbox`, `Commands`, `Files`, error hierarchy, and HTTP helpers. Also migrates docs from Fern to Mintlify, consolidates Biome config to a single root `biome.json`, and updates all E2E tests and documentation to match the new API surface.

Key changes:
- **TypeScript SDK**: New `Sandbox` class with static factory methods (`create`, `connect`, `list`, `get`, `killById`) and instance methods (`kill`, `pause`, `resume`, `update`, `waitForReady`), plus `Commands` and `Files` sub-modules
- **Python SDK**: Matching `Sandbox` and `AsyncSandbox` classes with snake_case API parity
- **Docs**: Fern docs removed, new Mintlify `docs.json`, `introduction.mdx`, and unified `quickstart.mdx`
- **Biome**: Consolidated to a single root `biome.json` for Biome 2.x compatibility, removed nested configs
- **Bug fixes**: Truthiness checks replaced with proper `!== undefined` / `is not None` guards, HTTP timeout margin added for command execution

## Additional Context

Design doc at `spec/2026-04-16-sdk-from-scratch-design.md` and implementation plan at `spec/2026-04-16-sdk-from-scratch-plan.md`.

## Testing

- [ ] Tested locally
- [x] Added/updated tests
- [ ] All tests passing


pavitrabhalla (commented):
## Review

Solid PR overall — the hand-crafted approach is much more maintainable than the Fern-generated code. HTTP layers are well-constructed (retry logic, backoff, signal composition, idle-timeout SSE). A few things to address before merging:

### Must Fix

**`waitForReady` is in the PR description but not in the code**
The description lists it as an instance method but it's absent from both SDKs. Either implement it or remove it from the description.

**Python `datetime.now()` fallback creates a naïve datetime**
In `types.py`, both `Field(default_factory=datetime.now)` and the `else datetime.now()` in `to_sandbox_info` produce a naïve datetime. Should be `datetime.now(tz=timezone.utc)`. (Note: `_parse_iso8601` correctly handles the 'Z' suffix already — this is just the fallback path.)

**TypeScript `res.json()` throws `SyntaxError` on empty non-204 responses**
If the server returns 200/201 with an empty body, `request<T>` will throw an uncaught `SyntaxError`. Safer:
```typescript
const text = await res.text()
return (text ? JSON.parse(text) : undefined) as T
```

**TypeScript `createdAt` fallback silently uses `new Date()` when `created_at` is absent**
```typescript
createdAt: raw.created_at ? new Date(raw.created_at) : new Date()
```
If the API omits the field, this silently sets creation time to "now". Better to make `createdAt` optional or throw.

---

### Should Fix

**Python `Sandbox` has no context manager support (`__enter__`/`__exit__`)**
Without it, users may forget to call `kill()` in error paths and leak the internal `httpx.Client`. Low-effort to add and standard SDK practice.

**No clear error after `kill()` closes the HTTP client**
After `kill()` closes `_http_client`, subsequent method calls produce a raw httpx error instead of something like `SandboxError("Sandbox already deleted")`.

**SSE line parsing is slightly stricter than the spec**
Both SDKs use `line.startsWith("data: ")` (mandatory space). The SSE spec allows `data:` with no space. A `.trim()` after slicing would make it robust at no cost.

---

### Minor

- `SandboxStatus` is missing states from the lifecycle docs (`starting`, `idle`, `failed`). If the API can return these, the Python SDK will raise a `ValueError` and TypeScript will silently pass through an unknown string. Worth confirming the contract.
- `Sandbox.status` goes stale after `resume()` — `files` is rebuilt but `status` stays as the construction-time snapshot. Already documented, just a heads-up for users who check it post-resume.
- The `Uint8Array` copy in `files.ts` `toBody()` is only needed for `SharedArrayBuffer` — could skip the copy for the common case.
- Testing checkboxes ("Tested locally", "All tests passing") are unchecked — please confirm before merging given the scope.

nirnejak:
### Must Fix
✅ waitForReady — wasn’t in the code on purpose (create is synchronous now per the latest OpenAPI spec: “When the request returns successfully, the sandbox is ready to use”). Removing it from the PR description.
✅ Python UTC datetime fallback — removed. created_at is now required in SandboxInfo, and to_sandbox_info throws SandboxError("missing created_at") if the API omits it. No more fabricated naive datetimes.
✅ TS res.json() on empty body — fixed with the res.text() + conditional parse pattern you suggested. Added a unit test.
✅ TS createdAt silent fallback — fixed. toSandboxInfo now throws on missing created_at, matching the existing validation for id and status.


### Should Fix
🚫 Python context manager — we deliberately removed __enter__/__exit__ in an earlier iteration to keep the lifecycle explicit; users call sandbox.kill(). HTTP client leakage on error paths is real but mitigated by (a) the new guard below that surfaces a clear error if you try to use a killed sandbox, and (b) the httpx.Client will be GC’d when the Sandbox is (after trying a try/finally: sandbox.kill() pattern, which is standard Python). Happy to revisit if this hurts real usage.
✅ Clear error after kill — added _require_live() guard. Calls like get_info, pause, resume, update on a killed sandbox now raise SandboxError("Sandbox '<id>' has been deleted; create or connect to a new one."). kill() itself remains idempotent.
✅ SSE line parsing — changed to line.startsWith("data:") / startswith("data:") with a .trim() / .strip() on the slice. Now matches the SSE spec (optional leading space).

## Minor
🚫 SandboxStatus missing starting/idle/failed — those were in older spec drafts but not in the current API contract. The current enum: [active, paused, resuming] is all the backend returns (confirmed with backend team). Our docs already match this.
✅ Uint8Array copy in toBody() — good catch, left the comment in place explaining it’s a SharedArrayBuffer guard. The copy is a couple of instructions per write and I’d rather keep the simpler code path over branching by buffer type. Happy to optimize if this shows up in a profile.
✅ Testing checkboxes — will check after the new commits verify in CI.

meAmitPatil (approved):

PR #192, opened 2026-05-15
Console: Terminal improvements

nirnejak:
## Description
Major Terminal Improvements

## Testing
- [x] Tested staging
- [x] Added/updated tests
- [x] All tests passing

## Screenshots/Demo

Screenshot 2026-05-16 at 5 26 17 AM

---

<a href="https://stagereview.app/superserve-ai/superserve/pull/192">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>


pavitrabhalla:
## Review

### Overview

A substantial terminal upgrade: 5 new xterm addons (WebGL renderer, search, serialize, image, web-links, ligatures), scrollback persistence via `localStorage`, auto-reconnect with exponential backoff, in-terminal search overlay (Cmd+F), font-size zoom (Cmd+=/−/0), and a switch from `display:none` to `visibility:hidden` DOM-parking for inactive tabs. Well-scoped and mostly clean.

---

### Bug — Tab close re-saves the buffer after clearing it

**`terminal-tabs.tsx:174`** calls `clearTerminalBuffer` synchronously, but the `SandboxTerminal` unmount cleanup (`saveTerminalBuffer`) runs *after* the re-render that removes the tab. Net effect:

1. `clearTerminalBuffer(key)` — clears storage
2. State update → re-render → unmount fires
3. `saveTerminalBuffer(key, ...)` — re-writes buffer

The buffer a user explicitly closed now persists until the next page load. Fix: set a `"closing"` flag on the terminal ref before calling `closeTab`, and skip the serialize step in cleanup if it's set.

---

### Code Quality

**Good patterns:**
- `if (wsRef.current !== ws) return` in every WS event handler — correctly guards against stale closures from reconnects.
- `isMountedRef` checked in all async callbacks (RAF, setTimeout, dynamic import).
- `pagehide` serialization as a backstop for hard reloads.
- `sendResize` extracted and reused everywhere — eliminates the previous copy-pasted inline JSON.

**Minor issues:**

- `const isForegroundTerminal = isActive` (`terminal.tsx:1075`) is a pointless alias — just use `isActive` directly.
- The search close button uses a `✕` character. The design system uses `@phosphor-icons/react` with `weight="light"` consistently — should be `<X weight="light" />`.
- `searchRef.current?.clearDecorations()` should be called when the search overlay closes (Escape / ✕ click) — otherwise xterm's match highlights persist after the overlay is dismissed.

---

### Bundle Size Concern — `@xterm/addon-ligatures`

This addon's transitive deps (`font-ligatures → font-finder → get-system-fonts, opentype.js`) are Node.js system-font scanning libraries. In the browser these won't function, but they will be bundled. `opentype.js` alone is ~300 KB. Since ligatures are loaded via dynamic import with try/catch, failures are silently swallowed — which is correct — but the bundle cost is paid regardless. Worth evaluating whether the ligature enhancement justifies the weight.

---

### Architecture Notes

- **`suggestedRenderer` module-level var:** Intentional (matches VS Code behavior) and clearly commented. If one terminal's WebGL context dies, all future terminals in the page session fall back to DOM.
- **Image paste handler sending `\x16`:** Works for TUI apps (Codex, Claude Code, opencode), but in a plain shell Ctrl+V inserts a literal `^V`. Low priority since it only fires when the clipboard contains a file with no text content.

---

### Test Coverage

Good coverage on the new behaviors. Missing:
- Font-size zoom (Cmd+=/−/0) — `applyFontSize` is exercised nowhere.
- `pagehide` serialization path.
- Search clear-decorations-on-close (once that's fixed).

---

### Summary

**Must fix before merging:** The tab-close re-save bug (`clearTerminalBuffer` → unmount → `saveTerminalBuffer`). Everything else is non-blocking. The ligature bundle concern is worth a conscious decision but not a blocker.

meAmitPatil (approved):

PR #195, opened 2026-05-26
docs: add Claude Managed Agents integration guide

meAmitPatil:
## Summary
- Adds integration guide for using Superserve sandboxes as the self-hosted execution environment for Anthropic's Claude Managed Agents
- Covers template setup, orchestrator (polling pattern), in-sandbox runner, and session creation with both TypeScript and Python examples
- Adds the guide to the Agent Harnesses group in the Integration Guides nav tab


pavitrabhalla:
## Competitor comparison: Daytona's Claude Managed Agents guide

I compared our page against Daytona's equivalent guide ([daytona.io/docs/.../claude-managed-agents](https://www.daytona.io/docs/en/guides/claude/claude-managed-agents/)) to spot gaps.

### Where we're ahead

- **Dual-language code** — every example is TypeScript + Python. Daytona is Python-only.
- **Egress lockdown** — we show `allowOut`/`denyOut` network rules and warn against leaking `ANTHROPIC_API_KEY` into the sandbox. Daytona covers none of this.
- **Pause/resume for cost savings** — we explain the checkpointing mechanism. Daytona just lists it as a bullet in "Advantages."
- **Firecracker isolation** — we explain the microVM model. Daytona calls them "containerized environments."
- **"What you get" section** — concrete, differentiated benefits vs. a thinner advantages list.

### Gaps vs. Daytona

| Gap | Impact |
|-----|--------|
| **No event streaming example** | High — Daytona shows `sessions.events.stream()` + `events.send()` together so readers see the full send-and-listen loop. We link out to Anthropic docs. |
| **No webhook orchestration variant** | Medium — Daytona covers a FastAPI webhook receiver as an alternative to long-polling. We mention it in one sentence. Matters for production users who want push-based dispatch. |
| **No pre-prepared sandbox pattern** | Medium — Daytona covers attaching a pre-seeded sandbox to a session via metadata. Useful for agents that need custom data loaded before session start. |
| **Vague agent creation step** | Low-medium — we say "create your agent as you would for any Managed Agents setup," which may leave first-time users lost. |
| **No MCP servers section** | Low — we mention it in "How it works" but don't revisit it. Daytona has a dedicated paragraph clarifying that MCP tools route through Anthropic, not the sandbox. |
| **No reference implementation link** | Low — Daytona points to a complete working repo. |

### Recommendation

Highest-impact addition is an **event streaming code snippet** — it's the final piece of the end-to-end flow and currently just deferred. After that, a **webhook variant** for the orchestrator would cover the production use case Daytona explicitly addresses.

pavitrabhalla (approved):

docs/integrations/managed-agents/claude-managed-agents.mdx
```diff
@@ -0,0 +1,528 @@
...
+
+Tool dispatch depends on the tool type:
+
+- **Filesystem and shell tools** (`bash`, `read`, `write`, `edit`, `glob`, `grep`) execute inside your Superserve sandbox. The tool runner handles each call against the sandbox's filesystem and shell, posting results back to the session.
+- **Web tools** (`web_search`, `web_fetch`) and **MCP server tools** route through Anthropic's servers. The sandbox is not involved.
+
+Each session gets its own sandbox. Filesystem state persists across tool calls, and idle sandboxes can be paused and resumed on demand.
+
+<Tip>
+A Python [reference implementation](https://github.com/superserve-ai/superserve/tree/main/guides/claude/claude-managed-agents) with the orchestrator, runner, and setup scripts is available on GitHub.
```
pavitrabhalla: @meAmitPatil Seems like we have examples for both Python and Typescript. Why link just Python here?

Also, can we link these reference implementations at the bottom of the article?

meAmitPatil: I added the TS example later, forgot to update the copy. Updating

PR #205, opened 2026-06-02
Add commands.spawn() for full-duplex exec (TS + Python) + docs

meAmitPatil:
Adds `commands.spawn()` — a live, full-duplex command session over the data-plane `/exec/connect` WebSocket. Complements the one-shot `commands.run()` (HTTP/SSE); pairs with the server side in superserve-ai/sandbox#89.

`spawn()` returns a handle while the process runs: stream output, write to stdin, signal it, and `wait()` for the exit result. Transport is never exposed — users pick `run` vs `spawn`, the SDK picks the wire.

## What's included
- **TypeScript** — `sandbox.commands.spawn()` → `CommandSession` (binary-safe `stdin.write`/`close`, `kill`, `wait`, `close`, `await using`). Native `WebSocket`, zero new deps.
- **Python** — async-only `AsyncCommands.spawn()` → `AsyncCommandSession` (`async with`, awaitable methods). Adds the `websockets` dependency (lazy-imported). The sync `Commands.spawn()` raises with guidance to use `AsyncSandbox`.
- **Docs** — new `commands/sessions.mdx`, SDK-reference `spawn`/`CommandSession`, nav + cross-links.

## Notes
- Wire is byte-exact: channel-tagged binary frames (0 stdin / 1 stdout / 2 stderr); lifecycle + control as text JSON. Output callbacks deliver text via a boundary-safe streaming decoder.
- A failed dial resumes a paused sandbox (`activate`) and retries once.
- Tests cover both SDKs identically: framing, split-UTF-8 decode, stdin/kill control, close-before-finish, resume-retry, socket cleanup.

TS 152 tests / Python 198 tests green; tsc, oxlint, oxfmt, ruff, mypy clean.


pavitrabhalla:
Really solid PR — the framing protocol, streaming decoder, settle-guard, and resume-retry are all handled correctly, and the test coverage is thorough.

A few things worth a look before merge:

1. **Python sync `spawn()` return type** — it always raises but is annotated `-> "AsyncCommandSession"`. Should be `-> NoReturn`.

2. **Broad exception catch in `_dial_with_resume`** — any error (including programming bugs) triggers a `refresh_activate` call before re-raising. Narrowing to connection-level errors would be more precise.

3. **No guard on `stdin.write`/`close` after session close** — calling these after `close()` will throw from the WebSocket layer. A silent guard or early error would be friendlier on both sides.

4. **Error frame exit code** — when the server sends `{"error": "...", "finished": true}` with no `exit_code`, the result defaults to 0. Worth confirming with sandbox#89 whether that's intentional — if the process never started, 0 is misleading.

docs/commands/overview.mdx
```diff
@@ -10,15 +10,18 @@ Run any shell command inside a sandbox using `sandbox.commands.run()`.
 ```typescript TypeScript
 import { Sandbox } from "@superserve/sdk"
 
-const sandbox = await Sandbox.create({ name: "data-analyzer" })
+const sandbox = await Sandbox.create({
+  name: "data-analyzer",
+  fromTemplate: "superserve/python-3.11",
```
pavitrabhalla, line 15: @meAmitPatil did we ever add the details in docs about what packages are installed in each default template?

docs/commands/streaming.mdx
```diff
@@ -32,48 +32,17 @@ print(f"Build exited {result.exit_code}")
...
-
-def on_stdout(data: str) -> None:
-    print(data, end="")
-    log_lines.append(data)
-
-result = sandbox.commands.run("./long-job.sh", on_stdout=on_stdout)
-
-upload_logs("".join(log_lines))
-```
-</CodeGroup>
```
pavitrabhalla, line 79: why was this block deleted?

meAmitPatil: Felt redundant, the note above already says result keeps the full output even while streaming.
Want me to add it back ?

docs/commands/overview.mdx
meAmitPatil: @pavitrabhalla Yes - https://docs.superserve.ai/templates/overview

docs/commands/streaming.mdx
pavitrabhalla: Having an example couldn't hurt. I'd say we don't change docs unless we find something inaccurate/misleading/stale, or if its customer feedback.

meAmitPatil: Cool, added it back.

pavitrabhalla (approved):

PR #219, opened 2026-06-10
console: typed errorCode for blocked-email + signup-tailored error page

meAmitPatil:
Follow-up to PR #218. Two improvements based on review feedback:

- Replace fragile string-match in the page with a typed `errorCode: "blocked_email"` returned from the signup action. Action also logs the blocked attempt (with email) for visibility.
- Pass `?reason=signup_blocked` to the auth-code-error page so the copy and "Try Again" destination match the signup context instead of saying "sign in". OAuth callback also redirects with the same query param when the trigger fires.

### DIFF
--- a/apps/console/src/app/(auth)/auth/auth-code-error/page.tsx
+++ b/apps/console/src/app/(auth)/auth/auth-code-error/page.tsx
@@ -1,12 +1,29 @@
-import { WarningIcon } from "@phosphor-icons/react/dist/ssr"
+"use client"
+
+import { WarningIcon } from "@phosphor-icons/react"
 import { Button } from "@superserve/ui"
 import Image from "next/image"
 import Link from "next/link"
+import { useSearchParams } from "next/navigation"
+import { Suspense } from "react"
 
 import { CornerBrackets } from "@/components/corner-brackets"
 import { DitherBackground } from "@/components/dither-background"
 
-export default function AuthCodeErrorPage() {
+function AuthCodeErrorContent() {
+  const searchParams = useSearchParams()
+  const reason = searchParams.get("reason")
+  const isSignupBlocked = reason === "signup_blocked"
+
+  const heading = isSignupBlocked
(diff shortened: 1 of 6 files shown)


apps/console/src/app/(auth)/auth/auth-code-error/page.tsx
```diff
@@ -1,12 +1,29 @@
...
+function AuthCodeErrorContent() {
+  const searchParams = useSearchParams()
+  const reason = searchParams.get("reason")
+  const isSignupBlocked = reason === "signup_blocked"
+
+  const heading = isSignupBlocked
+    ? "Signup Not Available"
+    : "Authentication Error"
+  const message = isSignupBlocked
+    ? "Signup is not available for this email address. Contact support@superserve.ai if you believe this is in error."
```
pavitrabhalla: why have this special message?

pavitrabhalla:
Good follow-up, addresses both issues cleanly. Two minor things:

**Extract the shared string constant**

`action.ts` and `callback/route.ts` both independently check `.includes("database error saving new user")`. These will drift if the message ever changes. Worth pulling into a shared constant, e.g.:

```ts
// lib/auth/errors.ts
export const BLOCKED_TRIGGER_MESSAGE = "database error saving new user"
```

**`rounded-full` in the Suspense fallback**

```tsx
<div className="h-5 w-5 animate-spin rounded-full border-2 border-primary border-t-transparent" />
```

Project design language is sharp corners everywhere — this spinner technically violates that. Probably no one will ever see it, but flagging for consistency.

Otherwise good to merge.

apps/console/src/app/(auth)/auth/auth-code-error/page.tsx
meAmitPatil: Changing to "Something went wrong. Please try again."

pavitrabhalla (approved):

PR #221, opened 2026-06-16
console: download a sandbox directory as a zip

mohamedsobhi777:
## What

Lets a console user download an entire sandbox directory as a single ZIP from the **Files → Download** panel. This is the console half of SS-66.

## Changes

`apps/console/src/components/sandboxes/files-section.tsx` (`DownloadPanel`):
- Removed the client-side `endsWith("/")` rejection — a directory path is no longer blocked before the request (`isValidAbsolutePath` is kept).
- Always send `&format=zip`; the server decides file-vs-directory (a file streams as-is, a directory comes back zipped).
- Derive the saved filename from the response `Content-Disposition` (so a folder saves as `<dir>.zip` and a file keeps its name), falling back to the path basename when the header is absent. The header comes from the sandbox (which runs untrusted user code), so it is treated as an untrusted boundary value: the filename is reduced to a basename and `.`/`..` are rejected, so it can't steer the saved path.
- The directory `400` guard now shows a deploy-aware "not available yet" message — it only fires before the data-plane rollout — which removes the deploy-order dependency between the two PRs.

## Test plan

`files-section.test.tsx`:
- [x] Rewrote the two old directory tests: a directory path now downloads a zip, and saves as `<name>.zip` from `Content-Disposition`
- [x] The file-download test now asserts the request carries `format=zip`
- [x] A path-traversal `Content-Disposition` filename (`../../etc/passwd`) saves as the basename (`passwd`)
- [x] A directory `400` from an un-rolled-out data plane shows the deploy-aware "not available yet" message
- [x] Existing upload / error-path / 404 tests unchanged and green
- [x] Full console suite 245/245 · `tsc --noEmit` clean · oxlint / oxfmt clean

## Review feedback

**Gemini:**
- ✅ **Filename path traversal** (high) — addressed in `7b139a1`: `filenameFromContentDisposition` reduces the value to a basename and rejects `.`/`..`, with a regression test. Browsers already sanitize the `<a download>` attribute, so this is defense-in-depth at the trust boundary rather than a reachable OS-traversal hole.
- ⏸ **`URL`/`URLSearchParams` for `filesUrl`** (medium) — declined: every key/value is already `encodeURIComponent`-encoded (no injection or malformed-URL risk in this code), and switching would change the shared upload path's space encoding (`+` vs `%20`) for no correctness gain. Easy to apply if there's a repo-wide preference.

**@awille-ai:**
- ✅ **Deploy-gap message** — addressed in `fb7f2fe`: the directory `400` now reads "Directory download is not available for this sandbox yet…" instead of implying directories are unsupported, removing the deploy-order dependency. Test updated to assert it.

## Notes / deploy order

Pairs with the sandbox data-plane PR — **superserve-ai/sandbox#114** (boxd `format=zip` + proxy `Content-Disposition` exposure). The feature is fully live once that ships, but the two **can deploy in any order**: if the console ships first, a directory download shows "Directory download is not available for this sandbox yet" (single-file downloads keep working) until the data plane catches up. SDKs are unchanged.


apps/console/src/components/sandboxes/files-section.tsx
```diff
@@ -35,6 +43,17 @@ function fileNameFromPath(path: string): string {
...
+// Pull the filename out of a Content-Disposition header. boxd controls the
+// format (`attachment; filename="<name>"`), so a simple match is enough — no
+// RFC 5987 `filename*` handling is needed. Returns null when the header is
+// absent or unparseable.
+function filenameFromContentDisposition(header: string | null): string | null {
+  if (!header) return null
+  const match = /filename=(?:"([^"]*)"|([^;]+))/i.exec(header)
+  const raw = match?.[1] ?? match?.[2]
+  return raw?.trim() || null
+}
```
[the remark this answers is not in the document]
mohamedsobhi777: Hardened in 7b139a1: `filenameFromContentDisposition` now reduces the value to a basename (slices after the last `/` or `\`) and returns null on `.`/`..`/empty, falling back to the trusted path basename. Agreed the data plane is an untrusted boundary.

One nuance: browsers already sanitize the `<a download>` attribute (path components stripped, write confined to Downloads), so this is defense-in-depth rather than a reachable OS-traversal hole — but validating the boundary input is the right call regardless. Added a regression test (`"../../etc/passwd"` → saves as `passwd`).

apps/console/src/components/sandboxes/files-section.tsx
```diff
@@ -25,8 +25,16 @@ import { formatBytes } from "@/lib/sandbox-utils"
...
+  sandboxId: string,
+  path: string,
+  params?: Record<string, string>,
+): string {
+  let url = `https://boxd-${sandboxId}.${SANDBOX_HOST}/files?path=${encodeURIComponent(path)}`
+  for (const [key, value] of Object.entries(params ?? {})) {
+    url += `&${encodeURIComponent(key)}=${encodeURIComponent(value)}`
+  }
+  return url
 }
```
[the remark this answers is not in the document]
mohamedsobhi777: Keeping the manual builder here. The helper already `encodeURIComponent`s every key and value, so there's no injection or malformed-URL risk in this specific code — the "error-prone" concern is generic rather than something this code exhibits.

Switching to `URL`/`URLSearchParams` would also change the encoding on the shared upload path (`URLSearchParams` encodes spaces as `+` vs the current `%20`) for no correctness gain, so I'd rather not churn a working, tested helper. Glad to switch if there's a repo-wide preference for the `URL` API.

awille-ai:
if (res.status === 400 && /director/i.test(detail)) {
  throw new Error(
    "Directory download is not available for this sandbox yet. Try again later or download individual files.",
  )
}
instead of ""/path" is a directory — enter a path to a specific file inside it." since that's misleading if the backend is not deployed. then we don't have a dependency and can deploy in any order.

The test should also update to check for this error.

mohamedsobhi777:
Done in `fb7f2fe`: the directory `400` guard now shows "Directory download is not available for this sandbox yet. Try again later or download individual files.", and the test asserts it.

Agreed this drops the deploy-order dependency — I verified console-first is safe: single-file downloads keep working and a directory just shows the "not yet" message until the data plane (#114) ships. Updated the PR's deploy-order note to reflect that the two can ship in any order.

meAmitPatil:
@mohamedsobhi777 Should we also provide an dedicated API endpoint to download a folder or maybe change the current download files endpoint to support this. If yes can you also include the API, SDK and Docs changes in the same PR or as a follow up.

mohamedsobhi777:
> @mohamedsobhi777 Should we also provide an dedicated API endpoint to download a folder or maybe change the current download files endpoint to support this. If yes can you also include the API, SDK and Docs changes in the same PR or as a follow up.

Makes sense, I have deliberately designed this to work via the console UI only, but I think the more general approach (API, SDK, CLI) will be more powerful. I'll update this PR since to reflect that.

awille-ai:
**High: unbounded download buffering enables client/browser DoS**

The new directory download paths all buffer the entire response before writing or saving it: the console uses `res.blob()`, the TS SDK uses `res.arrayBuffer()`, Python uses `response.content`, and the CLI writes only after `downloadDir()` returns the full byte array.

Because this response comes from a hostile sandbox, a user can request a directory or file that produces a multi-GB response, or an effectively endless chunked response. SDK/CLI request timeouts help with idle hangs, but they do not cap total bytes, and the browser path does not appear to have a timeout at all. This can exhaust browser memory, Node/Python process memory, or disk/network resources before the client gets a chance to reject it.

Please add bounded streaming for download paths where possible and enforce a maximum compressed byte count. Ideally the data plane should also enforce server-side limits before bytes leave boxd, with client-side limits as defense in depth.

awille-ai:
**High: console directory downloads need timeout/cancellation**

`downloadEntry` starts a browser `fetch(...format=zip...)` and then waits for `res.blob()` with no `AbortController`, timeout, idle timeout, byte cap, or user-visible cancellation path. A hostile sandbox can keep the response open while trickling bytes, leaving the UI stuck in the downloading state and continuing to consume memory/network until the browser or user kills the tab.

Please add an abortable download path for the console, with a clear timeout or idle timeout and a cancel control. This should pair with the byte-limit fix so both slow-never-ending responses and oversized responses are bounded.

apps/console/src/components/sandboxes/files/file-row.tsx
```diff
@@ -0,0 +1,111 @@
...
+interface FileRowProps {
+  entry: DirEntry
+  onOpen: (entry: DirEntry) => void
+  onDownload: (entry: DirEntry) => void
+  /** Drop OS files onto a folder row; ignored for file rows. */
+  onDropIntoFolder: (entry: DirEntry, e: React.DragEvent) => void
+  isDownloading: boolean
+}
+
+function hasFiles(e: React.DragEvent): boolean {
```
pavitrabhalla, line 18: Duplicate of the hasFiles function in file-browser.tsx:34 — identical implementations. Could be extracted to a shared utility (e.g. lib/drag-utils.ts) to avoid them diverging.

apps/console/src/components/sandboxes/files/file-browser.tsx
```diff
@@ -0,0 +1,359 @@
...
+  type DirEntry,
+} from "@/lib/api/files"
+import { fileKeys } from "@/lib/api/query-keys"
+import type { SandboxResponse } from "@/lib/api/types"
+import { FILE_EVENTS } from "@/lib/posthog/events"
+
+import { FileBreadcrumb } from "./file-breadcrumb"
+import { FileRow } from "./file-row"
+
+const DEFAULT_PATH = "/home/user"
```
pavitrabhalla, line 32: Is /home/user an invariant across all sandbox templates? If a template uses a different working directory, the browser opens on a path that may be empty or missing. Worth confirming this is a hard guarantee or deriving it from the sandbox.

apps/console/src/components/sandboxes/files/file-row.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return Array.from(e.dataTransfer.types).includes("Files")
+}
+
+function formatModified(unixSeconds: number): string {
+  if (!unixSeconds) return ""
+  return new Date(unixSeconds * 1000).toLocaleDateString(undefined, {
```
pavitrabhalla, line 24: No year in the output — 'Jun 19' is ambiguous for files from a prior year. Consider including year: 'numeric' when the date is more than ~6 months old (same heuristic GitHub uses).

.github/workflows/cli-publish.yml
```diff
@@ -25,6 +25,18 @@ jobs:
...
 
+      - name: Pin workspace runtime deps to published versions
+        working-directory: packages/cli
+        run: |
+          set -euo pipefail
+          # npm publish does NOT translate `workspace:*` in this bun-managed repo,
+          # so a literal "@superserve/sdk": "workspace:*" would ship and break
+          # `npm i @superserve/cli`. Rewrite it to the SDK's concrete version.
+          # The matching @superserve/sdk release must already be on npm.
+          SDK_VERSION="$(jq -r .version ../sdk/package.json)"
```
pavitrabhalla, line 36: The comment says the SDK must already be on npm, but there's no guard enforcing it. If publish order is wrong, the CLI ships with an unresolvable dep. Consider adding: npm view @superserve/sdk@$SDK_VERSION --json || { echo 'SDK version not on npm'; exit 1; }

spec/2026-06-16-cloud-coding-agents-linear-backlog.md
```diff
@@ -0,0 +1,433 @@
+# Cloud Coding Agents — Linear Backlog (DRAFT for review)
```
pavitrabhalla, line 1: This file (and the two sibling research docs for cloud coding agents) appear unrelated to the directory-download / file browser feature. Were these intentionally bundled here, or meant for a separate PR?

docs/filesystem/read-write.mdx
```diff
@@ -67,6 +67,44 @@ text: str = sandbox.files.read_text("/app/config.json")
...
+Superserve API key (`--api-key` or the `SUPERSERVE_API_KEY` environment
+variable), independent of `superserve login`:
+
+```bash
+# Saves to results.zip (the directory basename)
+superserve sandbox download <sandbox-id> /app/results
+
+# Choose the output path explicitly
+superserve sandbox download <sandbox-id> /app/results -o ./out.zip
+```
```
meAmitPatil: Lets remove this section to be uniform with other sections as we do have a CLI yet. We can add it back when we have one.

pavitrabhalla:
**Summary of open blockers before merge:**

The two HIGH findings from @awille-ai (posted Jun 19) are unaddressed and should be resolved first:

1. **Unbounded download buffering** — all download paths buffer the full response into memory before writing (`res.blob()` in the console, `res.arrayBuffer()` in the TS SDK, `response.content` in Python, full byte array in the CLI). A hostile sandbox can send a multi-GB or endless response to exhaust client memory. Needs bounded streaming and/or a max byte cap.

2. **No timeout or cancellation on console downloads** — `downloadEntry` fetches with no `AbortController`, no timeout, no byte cap, and no cancel UI. A hostile sandbox can trickle bytes indefinitely, leaving the UI stuck. Needs an abortable fetch path with a clear timeout and a cancel control.

Everything else (Gemini findings, @awille-ai deploy-gap, @meAmitPatil SDK/CLI request) has been addressed. The remaining open comments (docs CLI section, my inline comments) are non-blocking.

meAmitPatil (approved):
Approving, make sure you test the sdk properly before merging.

PR #227, opened 2026-06-18
feat(console): admin team impersonation (read-only, staff-gated)

mohamedsobhi777:
## Related Issues
Relates to SS-64

## Description
Adds read-only, console-only admin team impersonation behind a staff-identity gate. A staff user can "act as" a team from a new admin teams page; the console derives an ephemeral per-admin API key, scopes all reads to the impersonated team, blocks writes, and forwards only GET/HEAD through the API proxy. Impersonation state lives in a signed cookie, every start/stop is written to the audit log, and a banner renders across the dashboard while active.

## Security review follow-up (commit 8661182)
A review found the read-only guarantee was **bypassable via the data plane**, plus several perf/correctness gaps. All fixed:

**🔴 Critical — data-plane bypass (fixed).** The proxy GET/HEAD gate and server-action guards do **not** cover the terminal WebSocket or file upload/download — those connect **directly** from the browser to `boxd-…` using the per-sandbox `access_token`, which is handed to the browser by `GET /sandboxes/{id}` (a read, allowed while impersonating). So an admin could open any active customer sandbox → terminal → root shell + arbitrary file writes **as the customer** (logging `activity`/`proxy_audit` under their `team_id`, breaking the "zero usage" guarantee).
- **Enforcement:** the proxy now redacts `access_token` from JSON responses while impersonating (`lib/api/redact.ts`), so the browser never receives a usable data-plane credential. Ships with adversarial proxy tests that fail without the fix.
- **UX / defense-in-depth:** the terminal page, files section, and sandbox hero are gated behind a new `useImpersonation()` context and show a clear read-only state instead of broken controls.

**Hardening:**
- Impersonation-key `api_key` upsert no longer fires on every proxied GET (in-process cache; refreshes only past half-TTL).
- `listAllTeamsAction` bounded with a `LIMIT` (was an unbounded full-table select).
- Impersonation banner now **fails safe** (still shows on a team-name lookup error) instead of silently hiding while the session is active.
- `stopImpersonationAction` revokes the key best-effort (a transient DB error no longer blocks exit).
- `IMPERSONATION_TTL_MINUTES` clamped to an 8h ceiling.
- PostHog audit event (`admin_impersonation_started/stopped`) is now actually emitted (design D7).
- Deduped a double `getImpersonationTeamId` per proxy request.

Design doc (§5/§7) updated: data-plane access is now **blocked**, not an accepted out-of-scope gap.

## Additional Context
No backend changes — impersonation is entirely console-side using an ephemeral, auto-revoked key. Design and plan live in `spec/2026-06-17-admin-impersonation-{design,plan}.md`; new env vars are documented in `apps/console/.env.example`.

## Testing
Unit tests cover the staff gate, impersonation resolution, ephemeral key derivation/revoke + refresh caching, teams actions, the proxy gate (GET allowed, writes blocked), and the new `access_token` redaction (adversarial — fails without the fix). Full console suite: **277 passing**; typecheck + oxlint/oxfmt clean.

- [ ] Tested locally (manual browser e2e in progress)
- [x] Added/updated tests
- [x] All tests passing (local: 277 unit)

## Screenshots/Demo
N/A


apps/console/src/lib/admin/impersonation.ts
```diff
@@ -0,0 +1,120 @@
+import crypto from "node:crypto"
+
+import type { User } from "@supabase/supabase-js"
+import { cookies } from "next/headers"
+
+import { isStaff } from "@/lib/admin/staff"
+import { getProxySecret } from "@/lib/api/proxy-auth"
```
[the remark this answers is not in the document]
mohamedsobhi777: Thanks — I checked the import graph and there isn't actually a cycle here:

- `impersonation.ts` → imports `getProxySecret` from `@/lib/api/proxy-auth`
- `proxy-auth.ts` → imports `ensureImpersonationKeyRow` from `@/lib/admin/impersonation-key` (not from `@/lib/admin/impersonation`)
- `impersonation-key.ts` → only imports from `@/lib/api/proxy-secret` and `@/lib/supabase/admin`

Nothing imports back into `impersonation.ts`, so there's no circular import and no module-init `undefined` risk.

That said, the suggested change is still a good cleanup for a different reason: `getProxySecret` actually lives in `proxy-secret.ts` and `proxy-auth.ts` just re-exports it, so importing straight from `@/lib/api/proxy-secret` (the leaf module) avoids pulling `proxy-auth` into this file's graph. I'll make that change.

Good catch on the duplicate `auth.getUser()` round-trip in the related comments — I'll thread the layout's already-fetched `user` through `ImpersonationBanner` into `getImpersonationContext(user?)` with a fallback fetch so the existing tests stay green.

apps/console/src/lib/admin/impersonation.ts
```diff
@@ -0,0 +1,137 @@
...
+    path: "/",
+    maxAge: Math.floor(impersonationTtlMs() / 1000),
+    ...(cookieDomain ? { domain: cookieDomain } : {}),
+  })
+}
+
+export async function clearImpersonationCookie(): Promise<void> {
+  const store = await cookies()
+  store.delete(IMPERSONATION_COOKIE)
+}
```
awille-ai, line 108: High: `setImpersonationCookie` writes the cookie with `path: "/"` and, when configured, `domain: NEXT_PUBLIC_COOKIE_DOMAIN`, but `clearImpersonationCookie()` deletes only by name. In production with a shared cookie domain like `.superserve.ai`, this can fail to remove the domain-scoped cookie, so clicking Exit may leave the staff user impersonating on the next request. Please clear with the same path/domain attributes, or set an expired cookie with matching options, and add a test for the `NEXT_PUBLIC_COOKIE_DOMAIN` branch.

apps/console/src/lib/admin/audit.ts
```diff
@@ -0,0 +1,44 @@
...
+  adminEmail: string
+  teamId: string
+  teamName?: string
+}
+
+/** Per-individual, non-blocking audit. Never throws. */
+export async function logImpersonationEvent(
+  e: ImpersonationEvent,
+): Promise<void> {
+  console.info(
```
awille-ai, line 17: Non-blocking: for this stopgap, console + best-effort PostHog/Slack is useful operational visibility, but it is not durable enough to be the long-term source of truth for customer-data access. I would follow up with an append-only audit record for impersonation start/stop containing admin id/email, team id, action, timestamp, and ideally request metadata. This does not need to block this PR if the team is comfortable with the temporary risk.

awille-ai (commented):
Non-blocking architecture review comments on the stopgap impersonation implementation.

mohamedsobhi777:
**Synced with `main` (merge `0105d21`) + resolved conflicts.**

Conflict was in `files-section.tsx` — `main` replaced the upload/download panels with the new `FileBrowser`. Resolved by re-applying the impersonation gate to the new structure (the FileBrowser still uploads/downloads directly to the data plane with `access_token`, so it must stay hidden while impersonating). Verified the only other data-plane surfaces are the terminal (already gated) and FileBrowser (now gated).

Post-merge: typecheck clean, **316 console unit tests pass**, lint clean, Vercel console+ui builds pass.

apps/console/src/lib/admin/impersonation.ts
mohamedsobhi777: Fixed in 5680fb0. `clearImpersonationCookie()` now expires the cookie with the same `path: "/"` and (when set) `domain: NEXT_PUBLIC_COOKIE_DOMAIN` via `set(..., { maxAge: 0 })`, mirroring `setImpersonationCookie`. Added two tests covering the `NEXT_PUBLIC_COOKIE_DOMAIN` set/unset branches (assert the expiry carries the matching path/domain and the cookie is actually gone).

apps/console/src/lib/admin/audit.ts
mohamedsobhi777: Keeping this as the documented fast-follow `admin_audit` table with admin id/email, team id, action, timestamp).

apps/console/src/app/(dashboard)/sandboxes/[sandbox_id]/terminal/page.tsx
```diff
@@ -56,7 +58,7 @@ export default function TerminalPage() {
     )
```
Comment (awille-ai) [inline apps/console/src/app/(dashboard)/sandboxes/[sandbox_id]/terminal/page.tsx:58]: Non-blocking: this keeps the terminal body from mounting while impersonating, which is the important security control. There is still a separate paused-sandbox Start button in the header above that can call `resumeMutation` while impersonating; the proxy rejects it, so this is not a bypass, but hiding that header Start button during impersonation would avoid a guaranteed 403 path.

apps/console/src/app/api/[...path]/route.ts
```diff
@@ -55,6 +58,27 @@ async function proxyRequest(
...
   }
 
+  const supabase = await createServerClient()
+  const {
+    data: { user },
+  } = await supabase.auth.getUser()
+  // Resolve once and reuse: the read-only gate, the key injection, and the
+  // response redaction all key off the same impersonation decision.
+  const impersonatedTeamId = await getImpersonationTeamId(user)
+  const impersonating = impersonatedTeamId !== null
```
Comment (awille-ai) [inline apps/console/src/app/api/[...path]/route.ts:68]: Non-blocking: because impersonation allows all GET/HEAD requests through the proxy, the stopgap depends on upstream GET/HEAD routes remaining side-effect-free and not returning write-capable credentials. The `access_token` redaction covers the main data-plane credential; I would keep the allowed read surface deliberately small and add regression coverage if future GET endpoints mint/refresh credentials, enqueue work, or expose raw secret material.

awille-ai (commented):
Follow-up review on the latest head. No blocking issues found; two small non-blocking notes.

awille-ai (approved):
Reviewed the current head with a focus on the read-only impersonation boundary (staff gate, signed cookie/session handling, proxy GET/HEAD enforcement, API-key scoping, access_token redaction, and direct data-plane UI paths). I don’t see any critical blockers.

The earlier high-risk cookie clear issue appears fixed by expiring the cookie with the same path/domain attributes, and the data-plane token redaction plus UI gating cover the main terminal/file-transfer bypass. Remaining concerns look non-blocking/fast-follow: keeping the GET/HEAD proxy surface narrow over time, adding a durable audit table beyond logs/PostHog/Slack, and hiding any remaining guaranteed-403 UI affordances while impersonating.

PR #234, opened 2026-06-22
mcp: add @superserve/mcp server with docs and design spec

mohamedsobhi777:
## Related Issues

Implements SS-84 (Superserve sandbox MCP server).

## Description

Adds `@superserve/mcp`, a stdio Model Context Protocol server that exposes Superserve sandbox primitives as 10 fine-grained tools (`sandbox_create`, `sandbox_exec`, `sandbox_files_read`/`write`/`list`, and lifecycle) so any MCP client — Claude, Cursor, VS Code, Windsurf — can create sandboxes and run commands in isolated Firecracker microVMs. It's a thin wrapper over `@superserve/sdk`, authenticated by a single `SUPERSERVE_API_KEY`, with per-call `sandbox_id` targeting and transparent auto-resume on exec, so the per-sandbox data-plane access token never reaches the model. Also ships the `integrations/mcp` docs page (plus its nav entry), the package README, and the SS-84 research/design spec.

## Additional Context

- `sandbox_files_list` uses a `find`/`ls` shim over `commands.run` because the SDK has no native `files.list` yet — a follow-up will add a data-plane endpoint and swap the implementation with no tool-schema change.
- Output is truncated for context (32 KiB stdout/stderr, 1 MiB file reads), and SDK errors map to actionable `isError` messages so the agent can self-correct.
- Still open (not in this PR): npm + MCP-registry publish, and verification in live clients.

## Testing

`bunx turbo run test --filter=@superserve/mcp` — 35 unit + in-memory integration tests pass (tool schemas, error mapping, directory-listing parser, file round-trips); the live exec round-trip skips cleanly without `SUPERSERVE_API_KEY`.

- [x] Tested locally
- [x] Added/updated tests
- [x] All tests passing


mohamedsobhi777:
### Added: hosted MCP server

This commit adds a **hosted Streamable HTTP** deployment of the server so users can connect without a local install — same 10 tools, account-scoped, **stateless**, authenticated by passing a `ss_live_` API key as a bearer token (the inbound key is the exact credential the SDK already uses, so the platform's per-team scoping applies unchanged — no new isolation code).

**New files (reuse the existing transport-agnostic `createServer` / `createSdkClient`):**
- `src/lib/httpAuth.ts` — shared bearer extraction + per-request `buildServerForKey`
- `src/http.ts` — `handleMcpRequest(Request) => Response` web-standard handler, exported as `@superserve/mcp/http` for the console route (Next.js/edge)
- `src/httpServer.ts` — runnable Node server (`superserve-mcp-http` bin): `POST /` (or `/mcp`) + `GET /health`
- `tests/http.test.ts` — auth/method gating + an initialize-over-HTTP round-trip

**Verification:** oxlint 0/0, `tsc --noEmit` clean, 50 unit tests pass, deterministic two-pass build (bins get a shebang, `http.js` doesn't), and the built Node server smoke-tested locally — `/health` 200, missing-auth → 401, GET → 405, `initialize` round-trips.

pavitrabhalla:
@mohamedsobhi777 don't think you should check-in the spec docs here

packages/mcp/src/tools/files.ts
```diff
@@ -0,0 +1,173 @@
...
+        destructiveHint: false,
+        idempotentHint: true,
+        openWorldHint: true,
+      },
+    },
+    async ({ sandbox_id, path, encoding: enc }) => {
+      try {
+        const bytes = await client.readFile(sandbox_id, path)
+        const truncated = bytes.byteLength > MAX_FILE_BYTES
+        const slice = truncated ? bytes.subarray(0, MAX_FILE_BYTES) : bytes
```
awille-ai: High: this reads the entire sandbox file into the MCP process before applying the 1 MiB MCP truncation cap. The SDK's `files.read()` defaults to a much larger download cap, so a hosted MCP request can still buffer a very large file and OOM the serverless/container process even though the model only receives a truncated result. This is especially easy to hit accidentally by reading logs, artifacts, or archives.

Please push the cap down to the SDK/data-plane boundary or change the tool contract to reject over-large reads before buffering. Ideally the SDK exposes a partial/ranged read so this tool can preserve the advertised “large files are truncated” behavior without accepting multi-GB memory pressure.

packages/mcp/src/tools/files.ts
```diff
@@ -0,0 +1,173 @@
...
+        await client.writeFile(sandbox_id, path, data)
+        const bytes =
+          typeof data === "string"
+            ? Buffer.byteLength(data, "utf8")
+            : data.byteLength
+        return toolOk(`wrote ${bytes} bytes to ${path}`, { path, bytes })
+      } catch (e) {
+        return toolError(formatSdkError(e))
+      }
+    },
```
awille-ai, line 183: Medium: `content` is unbounded here, and for the hosted HTTP MCP path the JSON-RPC body has already been parsed into memory before this tool can write to the sandbox. A large text/base64 write can therefore consume MCP host memory and then expand further during base64 decode.

Consider adding explicit request/tool size limits for `sandbox_files_write` and returning an actionable error for larger payloads, or requiring large transfers to use a streaming/upload path instead of inline JSON content.

awille-ai (commented):
Architecture/security review focused on the hosted MCP surface, sandbox command/file tools, and SDK output limits.

Overall the design is directionally sound: bearer-token tenancy is delegated to the existing Superserve API key scoping, the HTTP handler is stateless per request, command output now has a bounded SDK response read, and the directory-listing shell path is quoted and traversal-checked.

I found one high availability/scalability issue around file reads buffering far more than the MCP tool returns, plus one medium hardening gap for hosted request/write size limits. I am not requesting changes because these are not critical enough to block under the review guidance, but I am also not approving because there is a high finding.

packages/mcp/src/client.ts
```diff
@@ -0,0 +1,194 @@
...
+    // Read-only: resolved via list() so it never resumes a paused sandbox
+    // (the SDK has no static single-sandbox get, and connect() would activate).
+    async info(id) {
+      const xs = await Sandbox.list({ ...conn })
+      const found = xs.find((s) => s.id === id)
+      if (!found) throw new NotFoundError(`Sandbox ${id} not found`)
+      return found
+    },
+
+    async exec(id, command, opts) {
```
awille-ai, line 380: [high] `sandbox_files_read` truncates after `client.readFile()` has already downloaded the whole file. In the real client this calls `sb.files.read(path)` without a `maxBytes`, so the SDK default allows buffering up to 2 GiB before the MCP layer slices to `MAX_FILE_BYTES`. On the hosted MCP server, any caller with a valid API key can create or point at a large file and force the serverless process to allocate far more memory than it will return. Please push the cap down into the SDK read, or add a capped/partial read API, and add a unit test that a file over the MCP limit does not get fully buffered.

packages/mcp/src/http.ts
```diff
@@ -0,0 +1,86 @@
...
+ * 405 so clients do not hold an idle stream open.
+ */
+export async function handleMcpRequest(request: Request): Promise<Response> {
+  if (request.method !== "POST") return methodNotAllowed()
+
+  const apiKey = extractBearerToken(request.headers.get("authorization"))
+  if (!apiKey) return unauthorized()
+
+  const server = buildServerForKey(apiKey, resolveBaseUrl())
+  const transport = new WebStandardStreamableHTTPServerTransport({
```
awille-ai, line 95: [high] The hosted HTTP path authenticates before dispatch, but it does not bound the request body before handing it to the MCP transport. Since tools like `sandbox_files_write` accept arbitrary text/base64 content, an authenticated but untrusted caller can POST a very large JSON-RPC body and exhaust memory before Zod/tool validation runs. Please enforce a maximum request size at this handler/route layer (and mirror it in the standalone Node handler) and add tests for rejected oversized bodies.

packages/mcp/src/tools/exec.ts
```diff
@@ -0,0 +1,97 @@
...
+      inputSchema: {
+        sandbox_id: z
+          .string()
+          .describe(
+            "ID of the sandbox to run in (from sandbox_create or sandbox_list).",
+          ),
+        command: z
+          .string()
+          .describe('Shell command to execute, e.g. "python script.py".'),
+        cwd: z
```
awille-ai, line 46: [medium] `timeout_ms` is only constrained to be positive. For the hosted MCP server this lets a caller ask the server to hold a request open for minutes or hours, tying up serverless/concurrency capacity until an outer platform timeout wins. Consider enforcing a sane maximum command timeout here and covering the boundary in unit tests.

awille-ai (commented):
Security/scalability review focused on tenant isolation, resource bounds, edge cases, and unit coverage.

I found high-severity resource-exhaustion risks in the hosted MCP path, but I did not find evidence of a critical cross-tenant authorization break in the changed code: the server is stateless per request and builds the SDK client from the caller's bearer token. Per the review guidance, I am not approving because of the high findings, and I am not blocking with REQUEST_CHANGES because I do not see a critical finding.

Test coverage is good for tool registration, core lifecycle behavior, listing parsers, UTF-16 truncation, and SDK sync exec response caps. The missing coverage maps to the findings below: oversized HTTP request bodies and oversized sandbox file reads should be bounded and tested.

docs/integrations/mcp.mdx
```diff
@@ -0,0 +1,249 @@
...
+      "servers": {
+        "superserve": {
+          "type": "stdio",
+          "command": "npx",
+          "args": ["-y", "@superserve/mcp"],
+          "env": { "SUPERSERVE_API_KEY": "${input:superserve-key}" }
+        }
+      }
+    }
+    ```
```
awille-ai, line 83: no codex?

packages/mcp/src/server.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SERVER_NAME, SERVER_VERSION } from "./constants.js"
+import { registerExecTool } from "./tools/exec.js"
+import { registerFileTools } from "./tools/files.js"
+import { registerLifecycleTools } from "./tools/lifecycle.js"
+
+export function createServer(client: SandboxClient): McpServer {
+  const server = new McpServer({
+    name: SERVER_NAME,
+    version: SERVER_VERSION,
+  })
```
awille-ai: [medium] The MCP server does not advertise server-level `instructions` in its initialize metadata. Codex consumes that field as server-wide guidance, and this server has several cross-tool constraints that are easy for an agent to miss from individual tool descriptions alone: create/list before per-sandbox calls, `sandbox_exec` and file tools auto-resume, prefer `sandbox_pause` vs `sandbox_kill` depending on state preservation, use `sandbox_template_list` before `from_template`, and read large files via bounded slices rather than full reads.

Please add concise MCP `instructions` when constructing `McpServer` and cover it in the schema/initialize tests. Keep the first ~512 chars self-contained so Codex gets the important workflow and limit guidance while deciding which tool to call.

docs/integrations/mcp.mdx
```diff
@@ -0,0 +1,249 @@
...
+
+| Variable              | Required | Description                                                               |
+| --------------------- | -------- | ------------------------------------------------------------------------- |
+| `SUPERSERVE_API_KEY`  | Yes      | Your Superserve API key (starts with `ss_live_`).                         |
+| `SUPERSERVE_BASE_URL` | No       | Override the control-plane URL (defaults to `https://api.superserve.ai`). |
+
+## Behavior and limits
+
+- **Auto-resume.** `sandbox_exec` and the file tools transparently resume a paused sandbox, so agents never need to call `sandbox_resume` first. `sandbox_resume` exists only to warm a sandbox explicitly.
+- **Output is capped for context.** `sandbox_exec` truncates stdout and stderr to 32 KiB each; `sandbox_files_read` truncates at 1 MiB. A truncated result sets `truncated: true` and reports the original byte length.
```
awille-ai: [low] This no longer matches the implementation. `sandbox_files_read` now passes `MAX_FILE_BYTES` into the SDK and returns an error with slice guidance when the file exceeds 1 MiB; it does not return a truncated result with `truncated: true` or the original byte length. Please update this docs line so agents and users do not expect partial content when the tool will reject the read.

awille-ai (commented):
Architecture/security follow-up on the latest head. The earlier high resource-exhaustion findings appear addressed by request-body caps, SDK-bounded file reads, and command timeout/output caps. I did not find a critical tenant-isolation issue in the current diff, and I am leaving this as a non-blocking commented review rather than requesting changes.

I did find two follow-ups: a medium Codex/MCP usability and safety gap around missing server-level instructions, and a low documentation mismatch around file-read limits.

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -0,0 +1,345 @@
...
+          .optional()
+          .describe(
+            "Environment variables available to commands in the sandbox.",
+          ),
+      },
+      annotations: {
+        readOnlyHint: false,
+        destructiveHint: false,
+        idempotentHint: false,
+        openWorldHint: true,
```
awille-ai, line 167: [medium] The MCP surface only lists templates and lets `sandbox_create` consume an existing `from_template`; it does not expose template creation/build. Since sandbox vCPU/memory/disk are inherited from the template and cannot be overridden at sandbox creation, an MCP client cannot satisfy a request like “create a 4 vCPU sandbox” unless the user has already created a 4 vCPU template outside MCP.

Please add a template creation workflow to the MCP surface, or at least track it explicitly before launch. A minimal useful shape would be something like `sandbox_template_create` with `name`, `from`, `vcpu`, `memory_mib`, `disk_mib`, and build steps, plus a way to wait/list status until the template is ready before passing it to `sandbox_create`.

awille-ai (commented):
Follow-up finding from reviewing the 4 vCPU sandbox workflow: the current MCP tool surface can list and consume templates, but it cannot create them. Since sandbox VM shape is inherited from the template and cannot be overridden at sandbox creation, that means MCP clients cannot create a 4 vCPU sandbox unless the user has already created a 4 vCPU template outside MCP.

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -0,0 +1,345 @@
...
+          .describe(
+            "Template (prebuilt base image) name or ID to base the sandbox on. " +
+              "Call sandbox_template_list first to see what your team has — don't guess a name.",
+          ),
+        from_snapshot: z
+          .string()
+          .optional()
+          .describe("Snapshot ID to restore the sandbox from."),
+        timeout_seconds: z
+          .number()
```
awille-ai, line 128: [medium] `sandbox_create` supports raw `env_vars`, but the MCP surface does not support creating Superserve secrets or binding/injecting existing secrets at sandbox creation time. That pushes agents toward passing credentials as plaintext environment variables, which is the less safe path and conflicts with the docs' guidance to use `secrets` for credentials.

Please add secret-aware creation support before launch, or explicitly document this as unsupported. Ideally the MCP surface should include either `secrets` on `sandbox_create` for binding existing stored secrets, plus separate secret-management tools if creation is in scope, or a clear workflow that directs users to create/bind secrets outside MCP.

awille-ai (commented):
Follow-up finding on secret handling: the current MCP server exposes `env_vars` during sandbox creation, but it does not expose Superserve secret creation or binding/injection into sandboxes at creation time.

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -0,0 +1,345 @@
...
+      name,
+      from_template,
+      from_snapshot,
+      timeout_seconds,
+      metadata,
+      env_vars,
+    }) => {
+      try {
+        const s = await client.create({
+          name,
```
awille-ai, line 183: [medium] Manual MCP workflow testing found a missing preview/discovery path: I was able to add a listener for port 4000 in the sandbox, but I could not retrieve the corresponding preview URL through the MCP server. That leaves agents able to start or expose a service but unable to hand the user the URL to inspect it.

Please add an MCP tool or field that exposes listener/preview URL discovery for sandbox ports, or document the supported workaround. For agent workflows, the useful shape is something like `sandbox_port_list` / `sandbox_port_open` returning the public preview URL for a given `sandbox_id` and port.

awille-ai (commented):
Follow-up from manual MCP workflow testing: I was able to add a listener for port 4000 from the sandbox, but I could not retrieve the corresponding preview URL through the MCP server.

packages/mcp/src/tools/lifecycle.ts
awille-ai, line 128: [high] Please treat network rules and secrets as blockers for the MCP launch.

Blockers:
- `sandbox_create` does not accept `network` rules (`allowOut` / `denyOut`), and there is no `sandbox_update` equivalent to add or change network policy after creation. Agents can therefore create sandboxes with broad default egress but cannot lock them down through MCP.
- `sandbox_create` does not accept `secrets`, and there are no attach/detach secret tools. That pushes credential workflows toward raw `env_vars` instead of the safer Superserve secret binding model.

Nice-to-have follow-up items to document or track separately:
- Network log access (`getNetworkLog`) for auditing what a sandbox actually reached.
- Streaming exec output for long-running commands.
- Interactive command sessions (`spawn`, stdin, signals, wait/close).
- Large file and directory transfer workflows (`downloadDir`, archive/streaming upload/download) beyond the small inline read/write path.
- Fuller `sandbox_info` parity for fields like network config and bound secret metadata.
- Billing/pricing/provider discovery if agents are expected to plan cost or secret provider setup.

The v1 can still stay intentionally small, but the docs should explicitly call out any non-blocker omissions and point to follow-up PRs so users do not assume the MCP server covers the full sandbox SDK/API surface.

awille-ai (commented):
Follow-up on MCP API surface coverage: there are a few sandbox API capabilities not exposed by the MCP server yet. Treat network rules and secrets as launch blockers; the remaining gaps can be documented as follow-up PRs if they are intentionally out of scope for v1.

mohamedsobhi777:
Thanks for the thorough passes, @awille-ai. Pushed 37a5137 addressing all of it — tool surface is now 18 (was 10), no SDK changes (it already exposed these), 84 unit/integration tests pass.

## Blockers

**Network rules** — `sandbox_create` now accepts `allow_out`/`deny_out`; new `sandbox_update` changes egress (or metadata) after creation; new `sandbox_network_log` audits actual egress (host/verdict/bytes + credential-injected requests).

**Secrets** — `sandbox_create` accepts a `secrets` binding map (`{ ENV_VAR: secret_name }`); new `sandbox_attach_secret` / `sandbox_detach_secret` for post-creation binding; new `secret_list` for discovery (metadata only — never values). `env_vars` now carries an in-schema warning to bind credentials with `secrets` instead.

One deliberate call: **secret _creation_ is intentionally not an MCP tool** — it takes a raw credential value that would otherwise transit the agent's tool-call transcript, which undercuts the broker model. Only binding existing secrets is exposed; creation is documented as an SDK/console step (per your "or a clear workflow that directs users to create/bind secrets outside MCP"). Happy to add `secret_create` if you'd rather have full parity.

## Medium

**Server `instructions`** — `McpServer` now ships `instructions` with the cross-tool workflow + the 1 MiB read limit front-loaded into the first ~512 chars; covered by an initialize test asserting `client.getInstructions()`.

**Template creation** — new `sandbox_template_create` (`name`, `from`, `vcpu`, `memory_mib`, `disk_mib`, `run_steps`, `start_cmd`, `ready_cmd`). It's async: returns `building`, and the agent polls `sandbox_template_list` until `ready` before passing it as `from_template` (blocking on a multi-minute build would pin hosted concurrency). This is the path to a 4 vCPU sandbox.

**Preview URLs** — new `sandbox_preview_url(sandbox_id, port)` → `https://{port}-{id}.{sandbox_host}`. Live-verified end to end (started `python3 -m http.server` on a sandbox; the returned URL served its content; prod apex correctly 404s for a staging sandbox). Port is range-validated and the sandbox id is host-safe-validated to block hostname injection.

**Fuller `sandbox_info`** — now returns network rules + secret bindings (plus timeout).

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -0,0 +1,679 @@
...
+          ),
+        allow_out: stringArray()
+          .optional()
+          .describe(
+            "Egress allowlist (host patterns, e.g. 'api.github.com'). When set, only these hosts " +
+              "are reachable; everything else is denied.",
+          ),
+        deny_out: stringArray()
+          .optional()
+          .describe(
```
awille-ai, line 158: [high] This description teaches an unsafe network-policy model. The platform docs say `allowOut` alone does not make a strict allowlist; to deny everything except allowed destinations, callers must combine `allowOut` with `denyOut: ["0.0.0.0/0"]`. As written, an agent can pass only `allow_out: ["api.github.com"]` believing everything else is denied, while the sandbox may still have broad default egress.

Please make the tool contract match the platform semantics. Either require/auto-add a deny-all rule when `allow_out` is provided for “strict allowlist” behavior, or change the descriptions and examples to say `allow_out` adds exceptions and strict lockdown requires `deny_out: ["0.0.0.0/0"]`. The same issue appears in `sandbox_update`'s `allow_out` description.

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -0,0 +1,679 @@
...
+  )
+
+  defineTool<NetworkLogArgs>(
+    server,
+    "sandbox_network_log",
+    {
+      title: "Audit a sandbox's egress",
+      description:
+        "List the outbound network connections a sandbox made (newest first) — host, verdict (allowed/blocked), " +
+        "bytes, and credential-injected requests. Use it to audit what a sandbox actually reached and to verify " +
```
awille-ai, line 621: [medium] `sandbox_network_log` currently auto-resumes paused sandboxes because the real client calls `Sandbox.connect()` before `getNetworkLog()`. Network log inspection should be a read-only control-plane audit operation; resuming a paused VM just to read historical egress logs changes state and can create unexpected cost/latency.

Please fetch the `GET /sandboxes/{sandbox_id}/network` control-plane endpoint without activating the sandbox, or add an SDK/static method that does that. Once it is side-effect-free, this tool can also be marked `readOnlyHint: true`.

docs/integrations/mcp.mdx
```diff
@@ -0,0 +1,292 @@
...
+    }
+    ```
+  </Tab>
+  <Tab title="Codex">
+    Add to `~/.codex/config.toml`. Codex also reads the server's `instructions` for cross-tool workflow guidance:
+    ```toml
+    [mcp_servers.superserve]
+    command = "npx"
+    args = ["-y", "@superserve/mcp"]
+    env = { SUPERSERVE_API_KEY = "ss_live_xxxxxxxxxxxxxxxx" }
```
awille-ai: [low] This Codex example stores the raw Superserve API key directly in `~/.codex/config.toml`. Codex supports forwarding environment variables to stdio MCP servers with `env_vars`, so the safer example is:

```toml
[mcp_servers.superserve]
command = "npx"
args = ["-y", "@superserve/mcp"]
env_vars = ["SUPERSERVE_API_KEY"]
```

For the hosted endpoint, prefer `bearer_token_env_var = "SUPERSERVE_API_KEY"`. This keeps docs from normalizing checked-in or copy-pasted secrets in config files.

awille-ai (commented):
Follow-up review on the latest head. Most earlier comments now have corresponding implementation or docs updates: server instructions, bounded file reads/request bodies, timeout caps, template creation, secret binding, preview URL, network update/log, and file-read docs are present. I found a few new issues in the newly added network/docs surface.

mohamedsobhi777:
Thanks @awille-ai — all three addressed in 87e1475:

- **[high] egress semantics** — fixed. Tool descriptions, server `instructions`, and docs now state that `allow_out` only adds allowed destinations and a strict allowlist needs `allow_out` + `deny_out: ["0.0.0.0/0"]` (and that `deny_out` is CIDRs-only). My earlier "only these hosts are reachable" / "empty array denies all" wording was wrong.
- **[medium] `sandbox_network_log` resume** — fixed. Now a direct, non-resuming control-plane `GET /sandboxes/{id}/network` (API-key, 30s timeout) instead of `Sandbox.connect()`, and marked `readOnlyHint: true`. Went with a direct fetch rather than a new SDK static method to avoid coupling the MCP to an unpublished SDK version before launch.
- **[low] Codex key in config** — fixed. Codex examples (docs + README) now use `env_vars = ["SUPERSERVE_API_KEY"]`, with `bearer_token_env_var` noted for the hosted endpoint.

Tests/lint/typecheck/format green (added a no-resume test for `sandbox_network_log`).

docs/integrations/mcp.mdx
```diff
@@ -0,0 +1,293 @@
...
+| `sandbox_files_read`      | Read a file (UTF-8 text, or base64 for binary).                                                                   |
+| `sandbox_files_write`     | Create or overwrite a file. Parent directories are created automatically.                                         |
+| `sandbox_files_list`      | List a directory's entries (name, type, size, modification time).                                                 |
+| `sandbox_pause`           | Pause a sandbox; state is preserved.                                                                              |
+| `sandbox_resume`          | Resume a paused sandbox (usually unnecessary — exec auto-resumes).                                                |
+| `sandbox_kill`            | Permanently delete a sandbox.                                                                                     |
+| `sandbox_preview_url`     | Build the public URL for a listening port (unauthenticated — anything on that port is internet-exposed).          |
+| `sandbox_network_log`     | Audit a sandbox's outbound connections (host, verdict, bytes). Auto-resumes a paused sandbox.                     |
```
awille-ai: [low] This table still says `sandbox_network_log` auto-resumes a paused sandbox, but the implementation now uses a direct control-plane read and the tool description says it is read-only and does not resume. Please update this row to say it does not resume, and add `sandbox_network_log` to the read-only tools list below so the docs match the tool annotations.

awille-ai (commented):
Follow-up on the latest head: the prior high network-policy semantics issue and the network-log resume side effect appear addressed in code/tool descriptions. I found one remaining documentation mismatch.

packages/mcp/src/client.ts
```diff
@@ -0,0 +1,428 @@
...
+    // would require Sandbox.connect (which resumes); the published SDK has no
+    // non-resuming static equivalent. Trusted control-plane endpoint, API-key
+    // auth, bounded by `limit` and a request timeout.
+    async networkLog(id, opts) {
+      const base = config.baseUrl ?? DEFAULT_BASE_URL
+      const url = new URL(`${base}/sandboxes/${encodeURIComponent(id)}/network`)
+      if (opts.limit !== undefined)
+        url.searchParams.set("limit", String(opts.limit))
+      if (opts.verdict !== undefined)
+        url.searchParams.set("verdict", opts.verdict)
```
pavitrabhalla, line 339: `opts.before` and `opts.since` are declared in `NetworkLogInput` but never forwarded to the URL here — and neither is exposed in the tool arg schema, so they're unreachable dead surface. Wire them through both the URL and the tool schema, or drop them from the interface.

packages/mcp/src/client.ts
```diff
@@ -0,0 +1,428 @@
...
+        vcpu: t.vcpu,
+        memoryMib: t.memoryMib,
+        diskMib: t.diskMib,
+        createdAt: t.createdAt,
+      })
+    },
+
+    // Read-only: resolved via list() so it never resumes a paused sandbox
+    // (the SDK has no static single-sandbox get, and connect() would activate).
+    async info(id) {
```
pavitrabhalla, line 316: This does a full `list()` scan to find one sandbox by ID. Given `sandbox_info` is annotated `readOnlyHint: true` (clients may call it frequently), is there a reason a static single-get can't be added to the SDK? Also: if `list()` is paginated, could a sandbox exist but not appear here?

packages/mcp/src/client.ts
```diff
@@ -0,0 +1,428 @@
...
+      // resort is a pure-POSIX shell loop, parsed like find's output.
+      const posix = await sb.commands.run(buildFallbackCommand(path))
+      if (posix.exitCode === 0) return parseFindOutput(posix.stdout)
+      const detail = (found.stderr || ls.stderr || posix.stderr || "").trim()
+      throw new SandboxError(
+        `Could not list ${path}${detail ? `: ${detail}` : ""}`,
+      )
+    },
+
+    async pause(id) {
```
pavitrabhalla, line 425: `connect()` resumes a paused sandbox, so calling `sandbox_pause` on an already-paused sandbox will resume it (cold-start cost) and immediately re-pause. Same applies to `attachSecret`/`detachSecret`. Is there a control-plane-only path for these that avoids the round-trip?

docs/integrations/mcp.mdx
pavitrabhalla: `sandbox_network_log` intentionally avoids `connect()` (uses a direct API-key fetch) so it does *not* auto-resume — this is the right behavior but the table copy is wrong.

packages/mcp/src/http.ts
```diff
@@ -0,0 +1,115 @@
...
+    sessionIdGenerator: undefined, // stateless — no session affinity
+    enableJsonResponse: true, // buffered JSON response, not a long-lived SSE stream
+  })
+
+  try {
+    await server.connect(transport)
+    const response = await transport.handleRequest(request, {
```
pavitrabhalla, line 102: `readWebRequestBody` already drained `request.body`, so the original `Request` passed here has a consumed stream. This works today because the transport uses `parsedBody` instead of re-reading, but it's fragile to upstream transport changes. Wrapping the buffered bytes in a fresh `Request` would be safer.

packages/mcp/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "mcp:validate": "mcpjam server validate --command node --args dist/index.js -e SUPERSERVE_API_KEY=${SUPERSERVE_API_KEY:-ss_live_dummy}",
+    "mcp:tools": "mcpjam tools list --command node --args dist/index.js -e SUPERSERVE_API_KEY=${SUPERSERVE_API_KEY:-ss_live_dummy}"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.29.0",
+    "@superserve/sdk": "workspace:*",
+    "zod": "^3.25.0"
+  },
+  "devDependencies": {
+    "@mcpjam/cli": "^3.10.0",
```
pavitrabhalla, line 63: `@mcpjam/cli` drags in the full `@ai-sdk` tree (Anthropic, OpenAI, Bedrock, Google, Azure, Mistral, xAI) plus `posthog-node`. Is this used in CI tests, or only for local inspection? If only local, worth noting in a dev-setup doc rather than a versioned dep.

.github/workflows/mcp-publish.yml
```diff
@@ -0,0 +1,114 @@
...
+          jq --arg v "$VERSION" '.version = $v' package.json > package.json.tmp && mv package.json.tmp package.json
+          jq --arg v "$VERSION" '.version = $v | (.packages[] | select(.registryType=="npm") | .version) = $v' server.json > server.json.tmp && mv server.json.tmp server.json
+
+      # npm publish does NOT translate `workspace:*`, so a literal
+      # "@superserve/sdk": "workspace:*" would ship and break `npm i`. Pin it to
+      # the latest version PUBLISHED on npm (the workspace may be ahead of the
+      # registry; the MCP server only uses the SDK's stable public API, so a
+      # caret range on the latest published version is correct and decouples
+      # this release from an unreleased SDK bump).
+      - name: Pin workspace runtime deps to published versions
```
awille-ai, line 84: Medium: the MCP package now depends on SDK functionality that is explicitly called out in `client.ts` as requiring `@superserve/sdk >= 0.7.7` (`Sandbox.files.downloadDir`). This workflow rewrites the workspace dependency to whatever `npm view @superserve/sdk version` returns at tag time, and the comment above still assumes the MCP server only uses SDK APIs that are safe to decouple from an unreleased SDK bump. If someone tags/publishes MCP before `@superserve/sdk@0.7.7` is published, the package can build against the workspace SDK but publish with a runtime dependency that lacks `downloadDir`, making `sandbox_files_download_dir` fail for users. Please add a minimum-version guard here (or pin/sequence the SDK release explicitly in the runbook) so the release fails fast instead of shipping an incompatible package.

awille-ai (approved):
Approved. I re-reviewed the latest head (`619b94e`) with a security/scalability lens and did not find any High findings.

The new directory download path is present end-to-end: `sandbox_files_download_dir` is registered, calls the SDK `downloadDir` with the MCP response cap, returns a base64 ZIP in structured content, and has an MCP tool test covering the call path. CI is green on the current head.

I left one Medium release-hardening note below about ensuring the MCP publish workflow cannot publish against an SDK version older than the one that introduced `downloadDir`; that is worth fixing or documenting before the first tagged release, but I do not think it should block this PR if release sequencing is controlled.

.github/workflows/mcp-publish.yml
mohamedsobhi777: Addressed in 644b7ea — added the enforceable guard here rather than relying on the runbook.

The "Pin workspace runtime deps" step now fails fast, before build/publish, if the latest published `@superserve/sdk` is below the floor the server actually calls:

```bash
MIN_SDK_VERSION="0.7.7" # floor: Sandbox.files.downloadDir (see src/client.ts)
SDK_VERSION="$(npm view @superserve/sdk version)"
test -n "$SDK_VERSION"
if [ "$(printf '%s\n%s\n' "$MIN_SDK_VERSION" "$SDK_VERSION" | sort -V | head -n1)" != "$MIN_SDK_VERSION" ]; then
  echo "::error::Latest published @superserve/sdk is $SDK_VERSION but @superserve/mcp requires >= $MIN_SDK_VERSION (Sandbox.files.downloadDir). Publish @superserve/sdk >= $MIN_SDK_VERSION before tagging mcp-v*, then re-tag." >&2
  exit 1
fi
```

So a tag can never ship an `@superserve/mcp` whose pinned runtime SDK lacks `downloadDir`: if the floor isn't on npm yet, the publish job stops with an actionable error (publish the SDK first, then re-tag). `sort -V` keeps the comparison version-aware, so `0.7.10`/`0.10.0` are ordered correctly rather than as strings.

I also fixed the now-inaccurate "decoupled from an unreleased SDK bump" wording in the file header and on the step, and pointed the `downloadDir` comment in `client.ts` at `MIN_SDK_VERSION` so the floor stays in sync across both ends. Thanks for the catch.

PR #247, opened 2026-06-29
Add PR Superloop: an installable agent loop that reviews PRs from a Superserve sandbox

mohamedsobhi777:
## Summary

Adds **PR Superloop** — an open-source, installable agent loop that shepherds open pull requests. On each commit pushed to a PR, Claude Code runs headless inside a warm Superserve sandbox, reviews the diff against a calibrated rubric, runs the project's own tests as a verifier, posts **one** concise review, and escalates risky changes to a human. It proposes; a human merges — it never merges, force-pushes, or edits CI.

This PR bundles the loop example + installer, the publishable `@superserve/loops` package and its release workflow, and the docs-site page.

## Changes

**Loop example — `examples/loops/`**
- `pr-superloop/loop.ts` — the loop implementation (find sandbox → warm checkout → review tick → pause).
- `pr-superloop/skill/SKILL.md` — the calibrated review skill (rubric, merge bar, safety rules).
- `pr-superloop/workflow.yml` — the GitHub Actions template dropped into a target repo.
- `pr-superloop/README.md` — full install + run-from-source guide.
- `install/cli.ts` — the `bunx @superserve/loops add pr-superloop` installer (runs `claude setup-token`, stores Superserve secrets, sets the Actions secret, writes the one workflow file).
- `lib/run-loop.ts` — the stateless orchestrator that runs one tick on the CI runner.
- `README.md` + `assets/` — loops-family overview with the wiring / tick / prompt-vs-loop visuals.
- `tests/`, `e2e/scenario.ts`, `scripts/bench.ts`, `tsconfig.json`, `vitest.config.ts` — unit tests, an e2e scenario, a benchmark, and config.
- `package.json` — the `@superserve/loops` package manifest.

**Packaging / CI**
- `.github/workflows/loops-publish.yml` — publishes `@superserve/loops` to npm on `loops-v*` tags.
- root `package.json` + `bun.lock` — register the new `examples/loops` workspace.

**Docs site**
- `docs/integrations/loops/pr-superloop.mdx` — the PR Superloop integration page (→ `docs.superserve.ai`).
- `docs/docs.json` — new **Loops** nav group (tagged `NEW`).

**Incidental — unrelated to the loop, flagging for awareness (happy to split out if preferred)**
- `apps/console/src/components/date-range-filter.test.tsx` — pins the clock so the test is date-independent.
- `docs/docs.json` — also adds Teams endpoints to the API Reference nav.

## Test plan
- [ ] `bunx turbo run test --filter=@superserve/loops` — loop unit tests pass.
- [ ] `bunx @superserve/loops run pr-superloop --repo owner/name --dry-run` prints the resolved plan (no keys needed).
- [ ] One live tick against a scratch PR posts a single review and pauses the sandbox.
- [ ] `bun run docs:build` renders the new page and the Loops nav entry.
- [ ] Tagging `loops-v*` publishes `@superserve/loops` via `loops-publish.yml`.


examples/loops/lib/run-loop.ts
```diff
@@ -0,0 +1,139 @@
...
+    : await ops.connect(existing.id)
+
+  try {
+    if (bootstrapped) {
+      for (const [path, contents] of Object.entries(spec.uploads ?? {})) {
+        await box.files.write(path, contents)
+      }
+      // The expensive step — runs once, every warm tick skips it.
+      await box.commands.run(spec.setup, { onStdout: log, onStderr: log })
+    }
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263 — `runLoop` now checks the setup exit code.

One step past the literal suggestion: a bare `throw` still leaves the just-created box behind, so the next tick would rediscover it via `list`, skip setup, and run `iterate` against the broken box forever — the permanent-breakage you flagged. So on setup failure it now tears the half-built box down (`box.kill()`, best-effort) **before** throwing, so the next tick re-bootstraps from scratch. New test: `setup failure: tears the box down, skips iterate + pause, and rethrows`.

examples/loops/pr-babysitter/loop.ts
```diff
@@ -0,0 +1,237 @@
...
+  const intervalMs = parseDuration(getFlag(args, "--watch"), 15 * 60_000)
+  console.log(
+    `[pr-babysitter] watching ${repo} every ${intervalMs / 1000}s — Ctrl-C to stop`,
+  )
+  await tick()
+  setInterval(() => {
+    void tick().catch((err) =>
+      console.error(`[pr-babysitter] tick failed: ${err}`),
+    )
+  }, intervalMs)
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263. Replaced `setInterval` with a recursive `setTimeout` that schedules the next tick only after the current one settles (`.finally(scheduleNext)`), so ticks run strictly sequentially and never overlap on the shared persistent box.

examples/loops/lib/run-loop.ts
```diff
@@ -0,0 +1,139 @@
...
+      sandboxId: box.id,
+      bootstrapped,
+      exitCode: result.exitCode,
+      stdout: result.stdout,
+    }
+  } finally {
+    if (spec.pauseWhenDone ?? true) {
+      await box.pause()
+    }
+  }
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263. `box.pause()` in the `finally` is now best-effort — `.catch()` logs a warning instead of letting a pause failure mask the real error escaping the `try`. It's also skipped entirely when the box was already torn down on a setup failure.

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,237 @@
...
+      "Claude: set SUPERSERVE_CLAUDE_SECRET (name of a Superserve secret holding the OAuth token) " +
+        "or CLAUDE_CODE_OAUTH_TOKEN (the raw token from `claude setup-token`).",
+    )
+  }
+
+  if (env.SUPERSERVE_GITHUB_SECRET) {
+    secrets.GITHUB_TOKEN = env.SUPERSERVE_GITHUB_SECRET
+  } else if (env.GITHUB_TOKEN) {
+    envVars.GITHUB_TOKEN = env.GITHUB_TOKEN
+  } else {
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263. `resolveAuth` now accepts `GITHUB_TOKEN || GH_TOKEN`, matching the installer CLI's resolution (and the `gh` CLI's own variable). New test: `accepts GH_TOKEN as a GitHub fallback`.

examples/loops/pr-babysitter/loop.ts
```diff
@@ -0,0 +1,237 @@
...
+    'git config --global user.name "loop-engineering[bot]"',
+    'git config --global user.email "loops@superserve.ai"',
+    `gh repo clone "$TARGET_REPO" ${REPO_DIR}`,
+  ].join("\n")
+}
+
+/** Per-tick: refresh the warm checkout, run Claude Code headlessly on the skill. */
+function iterateScript(): string {
+  return [
+    "# Subscription billing: ensure no metered API key out-ranks CLAUDE_CODE_OAUTH_TOKEN.",
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263. Added `set -e` as the first line of `iterateScript()` (matching `setupScript()`), so a failed `cd`/`git fetch` aborts the tick instead of running `claude` in the wrong dir against stale state. Asserted in the buildSpec test.

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,416 @@
...
+    return undefined
+  }
+}
+
+function detectRepo(flag: string | undefined): string {
+  if (flag) return flag
+  const url = tryExec("git", ["config", "--get", "remote.origin.url"])
+  const m = url
+    ? /github\.com[:/]([^/]+)\/([^/.]+?)(?:\.git)?$/.exec(url)
+    : null
```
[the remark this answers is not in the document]
mohamedsobhi777: Fixed in 7a42263 — added an optional trailing slash `\/?` before the anchor. Verified `owner/repo`, `…/repo/`, `…/repo.git`, `…/repo.git/`, and the `git@…:owner/repo/` SSH form all resolve to `owner/repo`.

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,248 @@
...
+  } catch (err) {
+    if (!dryRun) {
+      console.error(`error: ${(err as Error).message}`)
+      process.exit(1)
+    }
+    auth = { secrets: {}, envVars: {} }
+  }
+
+  const spec = buildSpec({ repo, skill, auth })
+
```
awille-ai, line 440: SHOULD_FIX: `runLoop` returns the sandbox command's `exitCode`, but this path only logs it. In the `--once` GitHub Actions workflow, a failed `iterate` command would still let `main()` resolve and the workflow step exit 0, so a broken tick can look healthy operationally. Consider throwing when `result.exitCode !== 0` (or setting `process.exitCode`) and add a unit test for a non-zero `iterate` result so scheduled failures are visible.

awille-ai (commented):
Architecture/security review pass: I do not see a critical security issue or a clear scalability blocker in the latest head. The earlier setup-failure, overlapping watch tick, pause masking, GH_TOKEN, set -e, and trailing-slash cases appear to be addressed.

I left one non-blocking reliability/test-coverage comment: failed `iterate` runs currently appear to leave the orchestrator process green, which can hide broken scheduled ticks in GitHub Actions. I would fix that before relying on the workflow operationally, but I am not blocking the draft PR on it.

Summary of this PR: it adds an `examples/loops` workspace with a reusable Superserve sandbox loop runner, a `pr-babysitter` orchestrator that runs Claude Code in a warm persistent sandbox to review open PRs, an installer CLI that vendors the loop into a target repo and wires secrets/workflow setup, docs/workflow examples, a live benchmark script, and unit tests around cold/warm loop behavior, auth resolution, and setup-failure cleanup.

docs/integrations/loops/pr-superloop.mdx
```diff
@@ -0,0 +1,198 @@
+---
+title: "PR Superloop"
```
pavitrabhalla: I don't think this makes sense to be in integrations. We may want to have a new section in our docs along the lines of Guides / Examples, etc. Integrations currently shows how we integrate with other branded 
products in the ecosystem.

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,382 @@
...
+    `[pr-loop] watching ${repo} every ${intervalMs / 1000}s — Ctrl-C to stop`,
+  )
+  // Recursive setTimeout, NOT setInterval: schedule the next tick only after the
+  // current one settles. Ticks share one persistent box, so an overlapping run
+  // (a slow tick outlasting the interval) would race on the same git checkout.
+  //
+  // Watch mode logs-and-continues: one bad tick must not kill a long-running
+  // watcher, so the exit code is intentionally ignored here (the .catch below
+  // handles a thrown tick). Only the one-shot path above maps it to the exit.
+  await runTick(spec)
```
pavitrabhalla: The first tick (this `await runTick(spec)`) runs bare — an error propagates through `runLoopCli` and crashes the process. The scheduled ticks inside `scheduleNext` get a `.catch()`, so they survive the same transient error. Could the first tick get the same treatment to make watch mode uniformly resilient?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,512 @@
...
+        stdin.setRawMode(false)
+        process.exit(1)
+      } else if (code === 127 || code === 8) {
+        // Backspace
+        buf = buf.slice(0, -1)
+      } else {
+        buf += ch
+      }
+    }
+    stdin.on("data", onData)
```
pavitrabhalla, line 146: No `stdin.on('error')` handler here — if the terminal disconnects mid-prompt (pipe break, SSH drop) the Promise neither resolves nor rejects and the installer hangs indefinitely. Would a `stdin.once('error', () => fail(...))` close this?

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,382 @@
...
+    "# Make the pr-loop skill discoverable by Claude Code.",
+    'mkdir -p "$HOME/.claude/skills/pr-loop"',
+    `cp ${SKILL_UPLOAD_PATH} "$HOME/.claude/skills/pr-loop/SKILL.md"`,
+    "# Authenticate git to github.com via the bound GITHUB_TOKEN. A credential",
+    "# helper reads it at call time, so the token never lands in .gitconfig (works",
+    "# for a raw PAT and for a Superserve proxy token swapped at egress). `gh` API",
+    "# calls authenticate separately from the GITHUB_TOKEN env var.",
+    'git config --global user.name "loop-engineering[bot]"',
+    'git config --global user.email "loops@superserve.ai"',
+    "git config --global credential.helper '!f() { echo username=x-access-token; echo \"password=$GITHUB_TOKEN\"; }; f'",
```
pavitrabhalla, line 182: The credential helper echoes `$GITHUB_TOKEN` to stdout as part of git's credential protocol. Under normal use only git reads it, but if `GIT_TRACE_CURL` or `GIT_TRACE_PACKET` is set (debug session, injected git config), the token lands in the Actions log. Worth calling out as a known tradeoff in the safety docs, or is there appetite to switch to `git credential approve` to avoid stdout exposure?

docs/docs.json
```diff
@@ -222,6 +222,10 @@
         "anchor": "Cookbook",
         "icon": "utensils",
         "groups": [
+          {
+            "group": "Recipes",
+            "pages": ["cookbook/pr-loop", "cookbook/software-factory"]
```
pavitrabhalla: `software-factory` is still the 'coming soon' placeholder — is it intentional to surface it in the nav here, or should it stay hidden until the recipe is ready?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,560 @@
...
+    )
+  }
+  try {
+    execFileSync("git", ["add", rel], {
+      cwd: repoRoot,
+      stdio: ["ignore", "ignore", "pipe"],
+    })
+    execFileSync(
+      "git",
+      ["commit", "-m", "Add pr-loop GitHub Actions workflow"],
```
pavitrabhalla: [BLOCKING · High] This commit has no pathspec, so it commits the entire staged index — not just the workflow file added on line 343. If the user had unrelated changes staged when they run the installer, those get swept into a commit labeled 'Add pr-loop GitHub Actions workflow' and pushed. Could this scope to the file, e.g. `git commit <rel> -m ...`?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,560 @@
...
+    c.ok(`committed ${rel}`)
+  } catch (err) {
+    c.warn(
+      `could not commit automatically (${(err as Error).message.split("\n")[0]}).`,
+    )
+    manualFallback()
+    return
+  }
+  try {
+    execFileSync("git", ["push"], {
```
pavitrabhalla, line 551: [BLOCKING · High] Bare `git push` publishes every local commit ahead of upstream, not just the workflow commit — a user with WIP commits on the branch would publish them unintentionally. It also runs unconditionally: interactive installs auto-push to the repo with no y/N confirmation (only --dry-run opts out). Given a push is hard to reverse, should this prompt when not --yes, or go back to printing the commands?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,560 @@
...
+    c.ok(`would commit and push ${rel}`)
+    return
+  }
+  const manualFallback = (): void => {
+    c.info(
+      `git add ${rel} && git commit -m 'Add pr-loop GitHub Actions workflow' && git push`,
+    )
+  }
+  try {
+    execFileSync("git", ["add", rel], {
```
pavitrabhalla, line 482: [BLOCKING · Medium] This commits + pushes to whatever branch is currently checked out. The common case is installing from `main`, but the workflow only triggers on `pull_request` — so pushing it to main opens no PR and the loop never fires, even though the installer prints 'Each new commit to a PR is reviewed within seconds.' Is the assumption that the user is on a feature branch? Worth handling the on-default-branch case explicitly.

pavitrabhalla (commented):
Re-review of the new auto-commit/push installer behavior plus a deeper pass over the loop and support scripts. Severity labels on each comment.

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,382 @@
...
+    'git checkout -f -q -B "$DEFAULT_BRANCH" "origin/$DEFAULT_BRANCH"',
+    `claude -p "/pr-loop ${task}" \\`,
+    "  --permission-mode dontAsk \\",
+    // Allow gh/git + the common test runners so the verifier can actually run the
+    // project's checks inside the microVM (safe — it's an isolated box).
+    '  --allowedTools "Bash(gh *),Bash(git *),Bash(npm *),Bash(npx *),Bash(node *),Bash(bun *),Bash(pnpm *),Bash(yarn *),Bash(make *),Bash(pytest *),Bash(python3 *),Bash(go *),Bash(cargo *),Read,Edit,Write,Glob,Grep" \\',
+    // Hard-block the dangerous ops at the tool layer, beyond the skill's rules.
+    '  --disallowedTools "Bash(gh pr merge*),Bash(git push*)" \\',
+    "  --output-format json \\",
+    "  --max-turns 50",
```
pavitrabhalla, line 258: [Non-blocking · Medium] With `set -e` above, a non-zero exit from `claude -p` (hit --max-turns, or a transient Anthropic rate-limit) aborts the tick and turns the Action red — indistinguishable from the sandbox/proxy actually breaking. 'Review incomplete' and 'infrastructure failed' become the same signal. Worth distinguishing a clean-but-unfinished run from a hard failure?

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,382 @@
...
+    // remote-tracking refs (origin/*): the default-branch working tree — where the
+    // reviewer reads project context (CLAUDE.md/AGENTS.md, test & lint commands) —
+    // would otherwise stay pinned at the clone-time SHA and go stale as the default
+    // branch moves. So fetch, then hard-reset that checkout to the remote tip.
+    "git fetch --all --prune --quiet",
+    // Re-derive the default branch (guards a rename since clone); fall back to main.
+    // Best-effort — the lookup itself must never fail the tick.
+    "git remote set-head origin --auto >/dev/null 2>&1 || true",
+    `DEFAULT_BRANCH="$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's#^origin/##' || true)"`,
+    `DEFAULT_BRANCH="\${DEFAULT_BRANCH:-main}"`,
```
pavitrabhalla, line 222: [Non-blocking · Medium] The fallback hardcodes `main`. On a `master`-default repo, if `refs/remotes/origin/HEAD` is ever unresolvable (the `set-head --auto` above is swallowed by `|| true` on a network blip), DEFAULT_BRANCH becomes `main`, then `git checkout -f -B main origin/main` fails and set -e aborts every tick until origin/HEAD is repaired. Could this derive from `git ls-remote --symref origin HEAD` instead of assuming main?

examples/loops/e2e/scenario.ts
```diff
@@ -0,0 +1,318 @@
...
+  console.log(`    opened ${prUrl}`)
+
+  console.log(
+    "[3] running the loop (cold start → Claude Code reviews the PR) ...",
+  )
+  const skill = readFileSync(
+    new URL("../pr-loop/skill/SKILL.md", import.meta.url),
+    "utf8",
+  )
+  const result = await runLoop(
```
pavitrabhalla: [Non-blocking · Low] If runLoop throws here, teardown (killLoopBoxes + repo delete at ~303) never runs — it's on the happy path, not in a finally. Every failed e2e leaks a private GitHub repo and a running sandbox. Wrap steps 1–4 in try/finally?

examples/loops/e2e/scenario.ts
```diff
@@ -0,0 +1,318 @@
...
+  const lastBody = bodies.length > 0 ? bodies[bodies.length - 1] : undefined
+  const review = bodies.find((b) => b.includes("pr-loop")) ?? lastBody
+
+  console.log("\n========== REVIEW ==========")
+  console.log(review ?? "(no review/comment found — see the loop output above)")
+  console.log("============================\n")
+
+  if (review) {
+    const lc = review.toLowerCase()
+    const caughtBug = lc.includes("discount") || lc.includes("test")
```
pavitrabhalla: [Non-blocking · Low] caughtBug matches on 'discount' and 'test', both of which appear in the PR title and filename — so a review that never actually identified the logic bug can still report RESULT: PASS. Match on something more specific to the bug (e.g. 'price - percent' / 'percentage')?

examples/loops/scripts/bench.ts
```diff
@@ -0,0 +1,83 @@
...
+    `    warm tick in ${secs(tWarm)}s — ${check.stdout.trim().replace(/\n/g, " | ")}`,
+  )
+  await warm.pause()
+
+  console.log("\n=== HERO NUMBERS ===")
+  console.log(`cold start (create + setup): ${secs(tCreate + tSetup)}s`)
+  console.log(`warm tick   (resume + run):  ${secs(tWarm)}s`)
+  console.log(`warm is ${((tCreate + tSetup) / tWarm).toFixed(1)}x faster\n`)
+
+  await Sandbox.killById(box.id)
```
pavitrabhalla: [Non-blocking · Low] killById isn't in a finally, so any throw after create (setup timeout, pause, resume) leaks the box. Relatedly, line 61 does `Sandbox.connect(info.id)` where `const [info]` can be undefined if list returns empty (paused boxes excluded / eventual consistency) — that throws before cleanup. Guard the empty case and move the kill into finally?

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,396 @@
...
+    // Headless in an isolated microVM → bypassPermissions. An allowlist can't work
+    // here: dontAsk silently DENIES any command shape not on the list, and the model
+    // picks the shape at runtime — a narrow allowlist made reviews never post because
+    // Claude built the body with a `cat <<EOF` heredoc and diffed via `VAR=$(...) &&
+    // git diff`, neither of which matched `Bash(gh *)`/`Bash(git *)`. Arbitrary bash is
+    // safe ONLY because of the sandbox boundary (per Anthropic's own guidance). The
+    // load-bearing "propose, never merge/push" guarantee is the WORKFLOW TOKEN SCOPE
+    // (contents:read → both `git push` and `gh pr merge` fail at the GitHub API), NOT
+    // the in-box mode. The deny list + skill rules below are defense-in-depth.
+    "  --permission-mode bypassPermissions \\",
```
pavitrabhalla, line 247: [BLOCKING · High] This trades the allowlist for arbitrary bash and leans entirely on token scope for the 'never merge/push' guarantee. But the README (pr-loop, ~line 87) offers 'a classic token with `repo` scope' as an identity option, and resolveAuth accepts any token without checking scope. With a repo-scoped classic PAT, a prompt-injected PR can drive the agent to `git push --force` or `gh api -X PUT .../pulls/N/merge` — both succeed, and neither is stopped by the deny list (no-op under bypass, and evadable by command shape regardless). Could resolveAuth assert scope and fail closed on write-capable tokens, and/or the docs drop the classic-repo option?

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,396 @@
...
+    // Defense-in-depth only (the token scope above is the real block; bypass may not
+    // enforce this): keep merge/push off the menu so a prompt-injected tick can't try.
+    '  --disallowedTools "Bash(gh pr merge*),Bash(git push*)" \\',
```
pavitrabhalla: [BLOCKING · Medium] Even in the safe default (contents:read), the token has pull-requests:write and the agent has arbitrary bash. None of these are `git push`/`gh pr merge`, so the deny list never covers them: injected PR content can steer `gh pr review --approve` (satisfy branch protection), `gh pr close` (kill a competing PR), `gh pr edit --add-label`/`--remove-label` (flip labels that gate downstream automerge), or `gh pr edit --base` (retarget). 'Propose, never act' doesn't hold for the full PR surface — is that acceptable, or should the skill/token constrain PR mutations too?

examples/loops/pr-loop/loop.ts
```diff
@@ -0,0 +1,396 @@
...
+    "# Make the pr-loop skill discoverable by Claude Code.",
+    'mkdir -p "$HOME/.claude/skills/pr-loop"',
+    `cp ${SKILL_UPLOAD_PATH} "$HOME/.claude/skills/pr-loop/SKILL.md"`,
+    "# Authenticate git to github.com via the bound GITHUB_TOKEN. A credential",
+    "# helper reads it at call time, so the token never lands in .gitconfig (works",
+    "# for a raw PAT and for a Superserve proxy token swapped at egress). `gh` API",
+    "# calls authenticate separately from the GITHUB_TOKEN env var.",
+    'git config --global user.name "loop-engineering[bot]"',
+    'git config --global user.email "loops@superserve.ai"',
+    "git config --global credential.helper '!f() { echo username=x-access-token; echo \"password=$GITHUB_TOKEN\"; }; f'",
```
pavitrabhalla, line 182: [Non-blocking · Medium] The credential helper resolves $GITHUB_TOKEN from the box env, so under bypassPermissions an injected agent can `echo $GITHUB_TOKEN` and exfiltrate it. The Claude OAuth token gets the egress-swap so it never enters the box, but the GitHub token does — and for a classic-repo/org PAT that leaks write access to every repo it can reach. Worth noting this asymmetry, or minting a per-run repo-scoped token for the box?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,624 @@
...
+  const workflowPath = writeWorkflow(
+    repoRoot,
+    {
+      githubApp: flags.githubApp,
+      // PAT and App identity are mutually exclusive; App wins if both are given.
+      githubSecret: !flags.githubApp && patToken ? GITHUB_SECRET : undefined,
+    },
+    flags.dryRun,
+  )
+  commitAndPushWorkflow(repoRoot, workflowPath, flags.dryRun)
```
pavitrabhalla: [BLOCKING · Medium] With --github-app this pushes a live workflow referencing secrets.LOOP_APP_ID / LOOP_APP_PRIVATE_KEY, but those are only printed as manual instructions afterward (~line 592). On the next pull_request event, create-github-app-token@v1 gets empty inputs and the whole tick job fails — the user is left with an already-live, red workflow. The default and PAT paths avoid this because their credential exists at push time. Set the App secrets before pushing, or skip the auto-push in App mode?

examples/loops/install/cli.ts
```diff
@@ -0,0 +1,624 @@
...
+        // Backspace
+        buf = buf.slice(0, -1)
+      } else {
+        buf += ch
+      }
+    }
+    stdin.on("data", onData)
+    // Without this, a mid-prompt disconnect (pipe break, SSH drop) leaves the
+    // promise never settling and the installer hanging indefinitely.
+    stdin.once("error", () => fail("stdin closed while waiting for input"))
```
pavitrabhalla: [Non-blocking · Medium] This handler (added to fix the earlier hang) is never removed on the success path — the resolve() branch drops the 'data' listener but not this one. After the prompt returns, the install runs several seconds of secret rotation + commit/push; if stdin then emits an error (SSH drop, EPIPE on the paused fd), fail() → process.exit(1) fires with a misleading 'stdin closed while waiting for input' after work partially completed. Capture it as a named onError and `stdin.off('error', onError)` alongside the data listener on the resolve and Ctrl-C paths?

examples/loops/tests/install.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    expect(wf).toContain("SUPERSERVE_GITHUB_SECRET: loop-github-token")
+    // The built-in token is dropped when a PAT identity is chosen.
+    expect(wf).not.toContain("github.token")
+    // Permissions block is still least-privilege regardless of identity path.
+    expect(wf).toContain("pull-requests: write")
+  })
+
+  it("mints a GitHub App installation token for the branded-bot identity", () => {
+    const wf = buildWorkflow({ githubApp: true })
```
pavitrabhalla, line 79: [Non-blocking · Low] These assert substrings but nothing parses the YAML, so an indentation drift or dropped newline that joins the app-token step onto setup-bun would still pass. A `yaml.parse(wf)` (or a structural assertion on steps) would catch a real break in the generated workflow.

examples/loops/lib/run-loop.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    // Re-inject envVars on EVERY tick, not just at create. Rotating credentials
+    // (notably the GitHub Actions `github.token`, which is minted per-run and dies
+    // when that run's job ends) are bound into the box only at bootstrap; a warm
+    // resume would otherwise keep running against the first tick's now-expired
+    // token and 401 on every `gh`/git call. Passing them here overlays the current
+    // run's fresh values on the box env for this command.
+    const result = await box.commands.run(spec.iterate, {
+      env: spec.envVars,
```
pavitrabhalla, line 152: [Non-blocking · Low] This re-injects envVars each tick but not secrets. It correctly refreshes the raw-env github.token path (where the per-run token actually lives), but a token bound via SUPERSERVE_GITHUB_SECRET sits in spec.secrets, which is only passed at create() and never re-injected on warm resume. Fine for a stable PAT — but the 'rotating credentials' comment reads more general than the code delivers.

pavitrabhalla (commented):
Re-review of the bypassPermissions switch, --github-app mode, and per-tick env re-injection. Severity labels on each. (Prior round's commitAndPushWorkflow comments still apply — that code is unchanged.)

examples/loops/pr-loop/loop.ts
mohamedsobhi777: Fixed in 742408d — fail closed at both ends.

- New `assertTokenScope()` in `pr-loop/loop.ts`: classic PATs are introspected via the `x-oauth-scopes` header (a free `/rate_limit` call); any write-capable scope (`repo`, `workflow`, `delete_repo`, `write:*`, `admin:*`) throws. It runs on the raw-token path each tick (`runLoopCli`) and in the installer before a `--github-token` PAT is stored — the last point the raw value is in hand (as a Superserve secret it never is again). Per-run `ghs_` tokens (bounded by the workflow's permissions block) and fine-grained `github_pat_` tokens skip introspection — GitHub doesn't report scopes for them, and they're the documented least-privilege paths. `LOOP_ALLOW_WRITE_TOKEN=1` is the explicit opt-in escape hatch.
- Docs: the pr-loop README and the cookbook page no longer offer the classic `repo`-scope option and now state those tokens are refused.

Tests: fails closed on `repo` scope, accepts a read-only classic PAT, skips `ghs_`/`github_pat_`, honors the escape hatch, and doesn't block when introspection itself fails offline (every later API call would fail the same way).

examples/loops/pr-loop/loop.ts
mohamedsobhi777: Two parts in 742408d, plus one correction on the approve vector.

- Approve: with the default workflow token, `gh pr review --approve` is blocked by GitHub itself — the repo setting "Allow GitHub Actions to create and approve pull requests" defaults to **off**, so the loop's reviews can't satisfy branch protection unless a repo explicitly opted in. The README's Safety section now says to leave that setting off.
- Close/reopen/ready are never legitimate for the loop → added to `--disallowedTools`. `gh pr edit` and `gh pr review` can't be blanket-denied (posting the review and the two loop labels needs them), so SKILL.md now has hard rules: comment-only reviews (never approve / request changes), never close/reopen/retarget/mark ready-or-draft, never touch labels other than `ready-to-merge` / `needs-human`.
- The README now documents the residual surface honestly: within `pull-requests: write`, the deny list + skill rules are guardrails under bypassPermissions, not guarantees — the guarantee remains that the token can never merge or push. GitHub has no finer-grained permission that allows posting reviews but forbids closing/relabeling; every review bot with this permission shares that surface.

examples/loops/pr-loop/loop.ts
mohamedsobhi777: Documented in 742408d (README Safety → "Token exposure in the box"). The asymmetry is narrower than it looks: the default path's in-box value is the per-run Actions token — masked in Actions logs, invalidated when the job ends, and bounded by the `contents: read` / `pull-requests: write` block. A PAT installed via `--github-token` never enters the box at all: the installer stores it as a Superserve secret, so the box sees a stand-in and the real value is attached at egress (same treatment as the Claude token). The only raw-PAT-in-box case is the local-dev env-var fallback — the docs now steer that to `SUPERSERVE_GITHUB_SECRET`, and the new scope gate (thread above) keeps write-capable classic PATs out of that path entirely, which caps the blast radius of an in-box `echo $GITHUB_TOKEN`.

examples/loops/install/cli.ts
mohamedsobhi777: Fixed in 742408d. The installer now checks whether `LOOP_APP_ID` / `LOOP_APP_PRIVATE_KEY` are already visible to the repo — both the repo-level and org-level Actions secrets endpoints; any lookup failure counts as not-configured (fail closed). If they exist, App-mode installs commit + push as before. If not, the workflow is committed but NOT pushed: the summary prints the two `gh secret set` commands followed by "then publish with: git push", and the premature `gh workflow run` hint is suppressed. A workflow referencing missing secrets can no longer go live from the installer.

examples/loops/install/cli.ts
mohamedsobhi777: Fixed in 742408d. The handler is now a named `onError`, removed together with the data listener in a shared `detach()` on both the resolve and Ctrl-C paths — it only covers the window where the prompt is actually pending, so a stdin hiccup during the secret-rotation / commit / push steps that follow can no longer kill the install with that misleading message.

examples/loops/tests/install.test.ts
mohamedsobhi777: Fixed in 742408d. Added `yaml` as a devDep and a structural suite alongside the substring checks: every identity mode is parsed and asserted on (`on.pull_request.types`, the exact least-privilege `permissions` block, steps as an array); App mode additionally asserts the mint step is a separate FIRST step with `id: app-token` and both secret inputs, and that the run step's env receives `${{ steps.app-token.outputs.token }}`. An indentation drift that joins the app-token step onto setup-bun now fails the suite.

examples/loops/lib/run-loop.ts
mohamedsobhi777: Fixed in 742408d — the comment now scopes the claim: secret bindings (`spec.secrets`) attach by NAME at create and the platform swaps the secret's current value in at egress, so rotation is picked up automatically with no re-injection; only raw env values go stale, which is why only `envVars` are overlaid per tick.

awille-ai (changes requested):
Security/correctness review complete. I’m requesting changes because the installer still performs destructive or surprising repository mutations outside the scope of the workflow file.

## High: the installer commits unrelated staged changes

`commitAndPushWorkflow` stages the workflow path, but then runs:

```sh
git commit -m "Add pr-loop GitHub Actions workflow"
```

with no pathspec. Git commits the entire staged index, including anything the user had staged before running the installer. That can silently sweep unrelated work into an installer-generated commit and misattribute it under the workflow commit message.

Please scope the commit to the workflow path (for example, `git commit -- <rel>` with the appropriate message/options), or preserve and restore the pre-existing index rather than consuming it.

## High: unconditional bare `git push` publishes unrelated commits

After creating the workflow commit, the installer runs a bare `git push`. That publishes every local commit currently ahead of the configured upstream, not just the workflow change. A user who runs the installer on a branch with unpublished work can unintentionally publish all of it. Interactive installs also do this without a confirmation prompt.

Please avoid automatically pushing by default, or require explicit confirmation/`--yes` after showing the exact branch and commits that will be published. At minimum, detect whether the branch was already ahead before the installer commit and refuse to auto-push in that case. Creating a dedicated branch/PR would be safer still.

These are existing unresolved review findings and remain present on the current head. CI is green, and the loop’s sandboxing, token-scope checks, setup-failure cleanup, and least-privilege default workflow are otherwise thoughtful. I’m withholding approval because an installer must not silently commit or publish unrelated user work.

mohamedsobhi777:
@awille-ai Both findings addressed in 4223c80.

**Scoped commit**: the installer now commits with a pathspec (`git commit -m … -- <workflow-file>`), so it records only the workflow file; anything the user had staged stays staged, untouched. Regression test runs against a real throwaway repo: stage `unrelated.txt`, run the installer commit, assert the commit contains exactly the workflow file and `unrelated.txt` is still in the index.

**Gated push** — publishing is now decided by a pure `planPush()`:
- If the branch was already ahead of upstream **before** the installer's commit, auto-push is refused outright — even under `--yes` — and the manual `git push` is printed. The push can therefore only ever publish the installer's own commit.
- When in sync it still needs consent: interactively a y/N prompt naming the exact commit and `branch → upstream`; non-interactively only `--yes` counts as consent (no TTY and no `--yes` → skip + print the command).
- No upstream / detached HEAD → skip + print.

Tests: `planPush` unit cases (including ahead-refusal beating `--yes`) plus real-git integration tests asserting upstream does not move in the ahead case and advances by exactly one commit in the consented case. README + cookbook updated to describe the confirmation and refuse-if-ahead behavior.

examples/loops/install/cli.ts
mohamedsobhi777: Fixed in 4223c80 — the commit is now pathspec-scoped (`git commit -m 'Add pr-loop GitHub Actions workflow' -- <rel>`), so only the workflow file is recorded and any pre-staged user work stays staged. Real-git regression test: stage an unrelated file, run the installer commit, assert the commit contains exactly the workflow file and the unrelated file is still in the index.

examples/loops/install/cli.ts
mohamedsobhi777: Fixed in 4223c80. Auto-push now (a) is refused outright — even with `--yes` — if the branch was already ahead of upstream before the installer's commit, so a push can only ever publish the installer's own commit; (b) when in sync, asks y/N interactively (naming the exact commit and branch → upstream) and treats `--yes` as the only non-interactive consent; (c) prints the manual `git push` on every skipped, declined, or failed path. Covered by `planPush` unit tests plus real-git integration tests (upstream unchanged in the ahead case; advances by exactly one commit in the consented case).

examples/loops/install/cli.ts
mohamedsobhi777: Half fix, half pushback here. The "pushing it to main opens no PR and the loop never fires" part doesn't hold: `pull_request` workflows run from the PR's merge ref, which includes base-branch content — so a workflow on the default branch fires for every PR opened after it lands, and `workflow_dispatch` actually requires the file on the default branch. Installing from main is the correct durable location, and the printed "each new commit to a PR is reviewed" claim holds for every subsequent PR. Installing from a feature branch also works (that PR's merge ref contains the file) and becomes durable once the branch merges.

What 4223c80 does change: the push is now confirmed interactively and names the exact `branch → upstream` it will publish to, so there's no longer a silent "which branch was I on" surprise — and it refuses to push at all if the branch carries unpublished commits.

awille-ai (approved):
Re-reviewed the current head. The two previous blocking installer issues are fixed, and I did not find any remaining critical or high-severity findings, so I’m approving.

Verified fixes:
- the installer commit is pathspec-scoped to the generated workflow file, so unrelated staged changes remain staged and are not swept into the installer commit;
- the installer snapshots the branch’s ahead count before committing and refuses to auto-push if any unpublished commits already exist;
- when the branch was in sync, push requires an explicit interactive y/N confirmation, or `--yes` as non-interactive consent;
- skipped and failed push paths print the manual command;
- tests cover the push plan and real-git staged/ahead scenarios.

CI is green on the current head. Remaining open comments are either already addressed in code or non-blocking operational/documentation concerns.

PR #255, opened 2026-07-02
docs: merge SDK and API reference into one tab, add Cookbook tab

mohamedsobhi777:
## Summary

Restructures the docs site navigation. The former 4-tab layout (Documentation · SDK Reference · Integration Guides · API Reference) becomes a single-axis **anchors** layout: no top tab bar; four icon-labeled sections pinned at the top of the sidebar, selected via a layout comparison across five rendered variants (#257–#260).

- **Documentation** — content and grouping unchanged.
- **SDK Reference** — the 5 existing pages, unchanged URLs.
- **REST API** — the OpenAPI-generated reference, moved as-is.
- **Cookbook** — new section holding the former Integration Guides: MCP, Coding Agents, Personal Agents, Managed Agents, and Agent Harnesses. All integration page URLs unchanged. A "Recipes" group (PR Loop, Software Factory placeholders) exists on disk under `docs/cookbook/` but is **hidden from the nav** until real content is ready — the pages aren't deleted, just not listed in `docs.json`.
- `scripts/check-docs-nav.ts`: the nav-sync check is now division-agnostic — it walks the whole navigation tree for endpoint entries and the `openapi` spec URL instead of assuming a tabs layout.
- `package.json`: `docs:build` and `docs:dev` were broken with the current Mintlify CLI (`build` command removed, `--dir` no longer honored); both now `cd docs` and use `mintlify validate` / `mintlify dev`.

## Test plan

- [x] `mintlify validate` passes (strict build validation)
- [x] `mintlify broken-links` passes
- [x] `bun run docs:check-nav` passes (44 operations in sync with the OpenAPI spec)
- [x] All 43 production `/api-reference/...` endpoint URLs (from the live sitemap) verified to still resolve 200 on the local preview — endpoint slugs are derived from the OpenAPI spec, not nav structure, so **no redirects needed**
- [x] Anchors layout rendered locally and screenshot-compared against tabs, dropdown, flat, and global-anchor variants
- [x] Confirmed Recipes group no longer renders in the Cookbook sidebar while `cookbook/pr-loop.mdx` and `cookbook/software-factory.mdx` remain on disk
- [ ] Verify the anchors sidebar renders as expected on the deployed Mintlify preview

PR #260, opened 2026-07-02
docs(layout): add global Console and GitHub anchors

mohamedsobhi777:
## Summary

Layout variant for comparison (stacked on #255). Keeps the current tab layout and adds **global anchors** — persistent Console and GitHub links pinned above the sidebar on every page, in every section.

Unlike the other three variants this is additive rather than an alternative shape, so it could also be combined with whichever layout wins.

Pure `docs/docs.json` change (+14 lines); no page files move and all URLs are unchanged.

## Test plan

- [x] `mintlify validate` passes
- [x] `bun run docs:check-nav` passes (44 operations)
- [x] Rendered locally and screenshotted (see the layout comparison in the workspace)

One of four alternative-layout PRs: anchors, dropdowns, flat sidebar, global anchors. Close the ones we don't pick.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -63,6 +63,20 @@
     }
   ],
   "navigation": {
+    "global": {
+      "anchors": [
+        {
+          "anchor": "Console",
+          "href": "https://console.superserve.ai",
+          "icon": "gauge"
+        },
+        {
+          "anchor": "GitHub",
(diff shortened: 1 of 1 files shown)


mohamedsobhi777:
Closing — the anchors layout (#257) was selected.

PR #261, opened 2026-07-02
sdk: derive API endpoint from region-tagged API keys

pavitrabhalla:
SDKs (TypeScript + Python) now derive the API endpoint client-side from the region token in `ss_live_<region>_<random>` API keys, via a known-regions map. This lets multi-region keys route to the right cell with zero configuration, part of the multi-region cell work.

**Technical decisions**
- Known-regions map instead of generic hostname derivation: an SDK must never construct a hostname before that region's DNS exists, so the map ships containing only live regions (`use` for now; `usw` lands once `api-usw.superserve.ai` / `usw-sandbox.superserve.ai` are up).
- Precedence, strictly: explicit `base_url`/`baseUrl` > `SUPERSERVE_BASE_URL` > region derived from the key > default. Sandbox host follows the same source.
- Legacy-key `_` collision is safe by design: legacy base64url randoms may contain `_`, so a legacy key can parse as having a "region" — any token not in the map falls back to the default endpoint, which is correct for every legacy key (all us-east). Region tokens are 1-17 lowercase alnum chars; parsing never throws.

**Testing**
- `packages/sdk`: `bun run test` (210 passed), `bun run typecheck`, `bun run lint` — all clean.
- `packages/python-sdk`: `uv run pytest` (281 passed), `uv run ruff check`, `uv run mypy` — all clean.
- New tests in both SDKs cover legacy keys, `use` keys, unknown regions, `_`-collision keys, weird/never-throw keys, and the full precedence order (plus prototype-member safety on the TS side via `Map`).

### DIFF
--- a/packages/sdk/src/config.ts
+++ b/packages/sdk/src/config.ts
@@ -10,6 +10,37 @@ import { AuthenticationError, ValidationError } from "./errors.js"
 const DEFAULT_BASE_URL = "https://api.superserve.ai"
 const DEFAULT_SANDBOX_HOST = "sandbox.superserve.ai"
 
+/**
+ * Known Superserve regions, keyed by the region token embedded in API keys
+ * (`ss_live_<region>_<random>`).
+ *
+ * Only regions whose DNS is live may be listed here: the SDK must never
+ * construct a hostname before that region's endpoints exist. `usw` will be
+ * added once `https://api-usw.superserve.ai` / `usw-sandbox.superserve.ai`
+ * are up.
+ *
+ * Legacy keys (`ss_live_<random>`, whose base64url random part may itself
+ * contain `_`) can coincidentally parse as having a region segment. That is
+ * harmless by design: any token not in this map falls back to the defaults,
+ * which is correct for every legacy key (they are all us-east).
+ */
+const KNOWN_REGIONS: ReadonlyMap<
+  string,
+  { baseUrl: string; sandboxHost: string }
+> = new Map([
+  [
+    "use",
+    {
+      baseUrl: "https://api.superserve.ai",
+      sandboxHost: "sandbox.superserve.ai",
+    },
+  ],
+])
+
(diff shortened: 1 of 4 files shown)


packages/sdk/src/config.ts
```diff
@@ -10,6 +10,39 @@ import { AuthenticationError, ValidationError } from "./errors.js"
...
+ * A region may be listed here only once its endpoints ACTUALLY SERVE THE
+ * API — DNS resolving is not the bar: superserve.ai has a catch-all
+ * wildcard, so every derived hostname resolves and answers with a
+ * valid-TLS 404 from the wrong infrastructure. Verify with a live /health
+ * check, not dig. `usw` will be added once `https://api-usw.superserve.ai`
+ * / `usw-sandbox.superserve.ai` serve their cell.
+ *
+ * Legacy keys (`ss_live_<random>`, whose base64url random part may itself
+ * contain `_`) can coincidentally parse as having a region segment. That is
+ * harmless by design: any token not in this map falls back to the defaults,
```
awille-ai: Medium, future-rollout concern: the unknown-token fallback protects legacy keys only while the parsed token is not in `KNOWN_REGIONS`. Once `usw` is added here, any legacy us-east key whose random part happens to start with `usw_` will be treated as a `usw` key and routed to the new cell, where it will fail auth even though the same key would work on the primary endpoint. The current map only has `use`, so this PR is safe today, but before adding non-`use` regions we should either make the key format unambiguous for newly region-tagged keys, exclude/migrate legacy keys with known-region prefixes, or document/accept the collision rate explicitly. Same concern applies to the mirrored Python map/comment.

awille-ai (approved):
Architecture review pass: I don’t see any critical or high-severity security/scalability issue in the current head.

The conservative known-region map is the right default for avoiding wildcard-DNS/catch-all failures, and the explicit/env override precedence remains intact. I left one medium rollout comment around legacy-key collisions once non-`use` regions are added to the map; it does not block this PR as written because only `use` is currently mapped.

PR #262, opened 2026-07-03
console: server-side pagination, sorting & filtering for sandboxes and templates

mohamedsobhi777:
## What

Adds server-side pagination, sorting, and filtering to the **Sandboxes** and **Templates** list views. Both pages previously fetched the entire list and filtered/searched in memory, which doesn't scale as a team accumulates sandboxes/templates. Each view now fetches only the current page and reads the total from the `X-Total-Count` response header.

## How

- **API client** (`lib/api/client.ts`): new `apiClientList` returns `{ items, total }`, reading `X-Total-Count` and falling back to page length when the header is absent — so it degrades gracefully against an API that doesn't paginate yet.
- **Hooks**: `useSandboxesPage` / `useTemplatesPage` take `{ page, pageSize, sort, order, status|owner, q }` and use `placeholderData: keepPreviousData` to avoid flicker. Optimistic mutation caches were reworked from a single full-array key to per-page `{ items, total }` caches (`setQueriesData` over `…lists()`).
- **URL state**: shared `useListParams` hook keeps page/size/sort/order/search in the URL (shareable, survives refresh); search is debounced.
- **Components**: new design-language `Pagination` (prev/next + numbered pages + "X–Y of Z" + page-size selector) and `SortableTableHead`.
- **Favicon**: `useFaviconStatus` (mounted globally in the dashboard shell) previously polled the **full** sandbox list every 10s — now a cheap `status=resuming&limit=1` probe.

## Backward compatibility

The API contract stays a bare array; pagination is **opt-in** query params + a response header. This PR degrades gracefully: against an API without pagination it shows one page of the full list, and fully activates once the paired backend PR is deployed.

**Depends on** the backend PR in `superserve-ai/sandbox` for the server-side params + `X-Total-Count` to take effect. (Backward-compatible, so ordering isn't strict.)

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run test` — 348 passing (incl. new `Pagination` tests + updated `use-sandboxes` hook tests)
- [x] `bunx oxlint` / `oxfmt --check` — clean
- [ ] Manual: paginate / sort / filter / search on both pages against a backend that has the pagination params (staging, post backend deploy)

### DIFF
--- a/apps/console/src/app/(dashboard)/sandboxes/page.tsx
+++ b/apps/console/src/app/(dashboard)/sandboxes/page.tsx
@@ -46,7 +46,7 @@ import {
   useSandboxesPage,
 } from "@/hooks/use-sandboxes"
 import { useSelection } from "@/hooks/use-selection"
-import type { SandboxListParams, SandboxSortColumn } from "@/lib/api/types"
+import { SANDBOX_SORT_COLUMNS, type SandboxListParams } from "@/lib/api/types"
 import { SANDBOX_EVENTS } from "@/lib/posthog/events"
 
 const STATUS_TABS = [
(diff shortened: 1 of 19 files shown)


mohamedsobhi777:
Paired backend PR: superserve-ai/sandbox#175 (adds the server-side pagination params + `X-Total-Count`).

mohamedsobhi777:
Addressed all five findings in 28ba922:

- **Out-of-bounds page after deleting the last row on the last page** (high, both pages): added an effect that snaps back to the last valid page when `page > pageCount` (guarded on `total > 0`).
- **Stale off-page selection** (high, sandboxes): now clears the selection when page / status / search changes. Note: I implemented this slightly differently than suggested — putting `clearSelection` (a fresh closure each render) in the effect deps would loop forever (`new Set()` is a new reference every run). Fixed properly by memoizing `clearSelection` with `useCallback` in `use-selection.ts`, so the effect only fires on actual view changes.
- **`router.replace` relative query string** (medium): now builds the full `${pathname}?${qs}`, consistent with the empty-query branch.
- **Missing `useEffect` import** (medium, templates): added.

typecheck + 348 tests + oxlint/oxfmt all green.

pavitrabhalla (approved):

PR #263, opened 2026-07-04
Reviving a page I killed

awille-ai:
## Related Issues
https://github.com/superserve-ai/superserve/pull/227 this was the original change that made the changes, and then i stomped on it by accident

## Description
This is the "impersonation" flow that I accidentally killed

## Testing
Added tests. also looking at the console build

- [X ] Tested locally
- [X ] Added/updated tests
- [X ] All tests passing


apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts
```diff
@@ -1,13 +1,65 @@
...
+import { createAdminClient } from "@/lib/supabase/admin"
 import { createServerClient } from "@/lib/supabase/server"
 
 const SANDBOX_API_URL =
   process.env.SANDBOX_API_URL ?? "https://api.superserve.ai"
+const PLATFORM_READ_KEY_NAME = "__console_platform_sandbox_read__"
+const PLATFORM_READ_KEY_VERSION = "v1"
+const DEFAULT_PLATFORM_READ_TTL_MINUTES = 30
+
+function platformReadTtlMinutes(): number {
```
Comment (pavitrabhalla) [inline apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts]: No `MAX_TTL_MINUTES` cap here — `impersonationTtlMs()` applies `Math.min(raw, 480)` on the same env var but this returns `raw` directly. With `IMPERSONATION_TTL_MINUTES=99999`, impersonation cookies cap at 8h but platform read keys get a ~69-day TTL written to `api_key`.

apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts
```diff
@@ -1,13 +1,65 @@
...
+  const raw = Number(process.env.IMPERSONATION_TTL_MINUTES)
+  return Number.isFinite(raw) && raw > 0
+    ? raw
+    : DEFAULT_PLATFORM_READ_TTL_MINUTES
+}
+
+function derivePlatformReadKey(adminId: string, teamId: string): string {
```
Comment (pavitrabhalla) [inline apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts]: `deriveImpersonationKey` in `impersonation-key.ts` does the same thing with a different prefix. Worth extracting a shared `deriveConsoleKey(purpose, adminId, teamId)` before this pattern gets copied a third time.

apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts
```diff
@@ -1,13 +1,65 @@
...
+  const mac = crypto
+    .createHmac("sha256", getProxySecret())
+    .update(
+      `${PLATFORM_READ_KEY_VERSION}:platform-sandbox-read:${adminId}:${teamId}`,
+    )
+    .digest()
+  return `ss_live_${mac.toString("base64url")}`
+}
+
+async function ensurePlatformReadKey(
```
Comment (pavitrabhalla) [inline apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts]: `ensureImpersonationKeyRow` skips the upsert when the cached expiry has >50% TTL left — this does not. Every admin sandbox page view unconditionally writes to `api_key`. The same `keyExpiryCache` pattern from `impersonation-key.ts` would fix it.

apps/console/src/lib/admin/teams-actions.ts
```diff
@@ -0,0 +1,102 @@
...
+    .eq("id", teamId)
+    .single()
+  if (error || !team) throw new Error("Team not found")
+
+  await setImpersonationCookie(teamId)
+  redirect("/sandboxes")
+}
+
+export async function stopImpersonationAction() {
+  const user = await requirePlatformSandboxRead()
```
pavitrabhalla: If the session expires while impersonating, this throws before `clearImpersonationCookie()` runs — the cookie stays set. On next login the user silently resumes impersonating the same team. Could clear the cookie first, then check permissions.

apps/console/src/app/(dashboard)/admin/admin-teams-table.tsx
```diff
@@ -0,0 +1,73 @@
...
+                  <TableCell className="text-muted tabular-nums">
+                    {team.created_at
+                      ? formatDate(new Date(team.created_at))
+                      : "—"}
+                  </TableCell>
+                  <TableCell className="text-right">
+                    <form action={onActAs.bind(null, team.id)}>
+                      <button
+                        type="submit"
+                        className="inline-flex h-8 items-center justify-center border border-border px-3 text-sm text-foreground transition-colors hover:bg-muted/30"
```
pavitrabhalla: Missing `border-dashed` and wrong button typography — should be `border border-dashed border-border font-mono uppercase text-xs`.

apps/console/src/app/(dashboard)/admin/teams/[team_id]/sandboxes/[sandbox_id]/sandbox-detail-client.tsx
```diff
@@ -0,0 +1,269 @@
...
+  },
+  failed: {
+    label: "Failed",
+    bg: "bg-destructive/[0.04]",
+    dot: "bg-destructive",
+    pulse: false,
+  },
+} as const
+
+function formatTimeout(seconds?: number | null): string {
```
Comment (pavitrabhalla) [inline apps/console/src/app/(dashboard)/admin/teams/[team_id]/sandboxes/[sandbox_id]/sandbox-detail-client.tsx]: Both already exist in `sandbox-resource-bar.tsx`, `sandbox-info-grid.tsx`, and a few template components — this is at least the 4th copy of each. `@/lib/format` has `formatTime`/`formatDate`; these belong there.

apps/console/src/app/api/[...path]/route.ts
```diff
@@ -56,6 +58,25 @@ async function proxyRequest(
     return NextResponse.json({ error: "Not found" }, { status: 404 })
   }
 
+  const supabase = await createServerClient()
+  const {
+    data: { user },
+  } = await supabase.auth.getUser()
```
Comment (pavitrabhalla) [inline apps/console/src/app/api/[...path]/route.ts]: `getUser()` was previously only called inside `!shouldSkipKeyInjection`. Now it runs for all allowed paths, including skip-injection ones — those now pay a Supabase round-trip they did not before. Also, the write-block below fires before the skip check, so a non-GET request to a skip-injection path during impersonation gets a 403 instead of forwarding.

apps/console/src/app/(dashboard)/admin/teams/[team_id]/sandboxes/[sandbox_id]/sandbox-detail-client.tsx
```diff
@@ -0,0 +1,257 @@
...
+  rows: Array<{ label: string; value: string }>
+}) {
+  if (rows.length === 0) {
+    return <p className="px-4 py-4 text-sm text-foreground/40">None</p>
+  }
+
+  return (
+    <div className="space-y-3 px-4 py-4">
+      {rows.map((row) => (
+        <div key={row.label} className="grid grid-cols-[140px_1fr] gap-4">
```
Comment (pavitrabhalla) [inline apps/console/src/app/(dashboard)/admin/teams/[team_id]/sandboxes/[sandbox_id]/sandbox-detail-client.tsx]: `networkRows` gives every `allow_out` the label `Allow` and every `deny_out` the label `Deny`, so two allow rules both render `key="Allow"`. Key on `row.value` or the index instead?

apps/console/src/app/(dashboard)/admin/teams/[team_id]/team-sandboxes-client.tsx
```diff
@@ -0,0 +1,137 @@
...
+          {sandbox.name}
+        </Link>
+      </TableCell>
+      <TableCell>
+        <Badge variant={statusVariant(sandbox.status)} dot>
+          {statusLabel(sandbox.status)}
+        </Badge>
+      </TableCell>
+      <TableCell className="font-mono text-xs text-muted tabular-nums">
+        {sandbox.vcpu_count}CPU | {sandbox.memory_mib}MB
```
Comment (pavitrabhalla) [inline apps/console/src/app/(dashboard)/admin/teams/[team_id]/team-sandboxes-client.tsx]: Detail view uses `formatMemory(sandbox.memory_mib)` (`2 GB`) but the list renders raw (`2048MB`). Use `formatMemory` here too for consistency?

apps/console/src/app/(dashboard)/layout.tsx
```diff
@@ -1,14 +1,23 @@
...
 import { QueryProvider } from "@/components/query-provider"
+import { getImpersonationContext } from "@/lib/admin/impersonation"
 
-export default function DashboardLayout({
+export default async function DashboardLayout({
   children,
 }: {
   children: React.ReactNode
 }) {
+  const impersonationContext = await getImpersonationContext()
```
pavitrabhalla: This awaits `getImpersonationContext()` (a `supabase.auth.getUser()` + cookie read) on every dashboard render for every user, not just admins. Is the round-trip request-memoized, or worth gating behind a cheaper cookie-presence check first?

apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts
```diff
@@ -73,16 +104,17 @@ async function proxyPlatformSandboxRead(
   }
 
   const { path = [] } = await params
-  const upstreamPath = targetPath(teamId, path)
+  const upstreamPath = targetPath(path)
   if (!upstreamPath) {
     return notFound()
   }
 
+  const apiKey = await ensurePlatformReadKey(user, teamId)
```
Comment (pavitrabhalla) [inline apps/console/src/app/api/platform/sandboxes/[[...path]]/route.ts]: `teamId` here is the raw `?team_id=` query value. `startImpersonationAction` runs `UUID_RE.test(teamId)` before touching `api_key`, but this path feeds it straight into `ensurePlatformReadKey` → `api_key.upsert`. Should we apply the same UUID guard so a `platform:sandbox:read` user can't write arbitrary `team_id` rows?

apps/console/src/lib/admin/teams-actions.ts
```diff
@@ -0,0 +1,122 @@
...
+  const { data: team, error } = await admin
+    .from("team")
+    .select("id")
+    .eq("id", teamId)
+    .single()
+  if (error || !team) throw new Error("Team not found")
+
+  await setImpersonationCookie(teamId)
+  redirect("/sandboxes")
+}
```
pavitrabhalla, line 80: This still drops the cookie clear when the auth check fails: if `requirePlatformImpersonationAccess()` throws (e.g. session expired while impersonating), the catch redirects to /admin without calling `clearImpersonationCookie()` — same behavior as before, just wrapped in try/catch now. On next login the user resumes impersonating the same team until the cookie's own TTL expires. Only the happy path (successful auth check) is covered by `teams-actions.test.ts`. Move the cookie clear into (or ahead of) the catch?

pavitrabhalla:
Re-reviewed after the latest push. Good rework — the platform-sandboxes proxy route got folded into the unified `/api/[...path]` proxy and most of what I flagged last round is resolved as a side effect:

- TTL cap: `getAuthApiKeyForUser` now derives the impersonation key TTL from `impersonationTtlMs()` itself (capped at 480min) instead of an uncapped raw env read.
- Duplicate key derivation: `deriveConsoleKey`/`deriveImpersonationKey` extracted and shared.
- Unconditional upsert: `ensureImpersonationKeyRow` now has the same `keyExpiryCache` >50%-TTL-left skip as the proxy key path.
- `getUser()`-for-every-path and write-block-before-skip-check: gone, since the new `route.ts` only touches auth/impersonation state inside the `!skipKeyInjection` branch.
- UUID guard on team_id: moot now — impersonated team_id comes from the signed cookie (set once, validated in `startImpersonationAction`), not a user-supplied param on every request.
- `getImpersonationContext` on every render: gated behind `hasImpersonationCookie()` first in layout.tsx.
- Crash findings (status fallback, network null-check), row key collision, formatMemory/formatTime duplication: all fixed, and the sandbox row now navigates via a real `<Link>` instead of an onClick div, which sidesteps the keyboard-a11y gap too.

Two things left one comment inline on the cookie-clear-on-throw issue in `teams-actions.ts` (still not fixed).

Also, flagging cross-repo: sandbox#195's `GetSandboxByID` still returns `secrets` (env_key/secret_name) unconditionally for `platform:sandbox:read` impersonation keys. `redactAccessTokens` here only strips `access_token`, not `secrets`, so if that API-side leak isn't fixed before both land, an admin viewing another team's sandbox detail page gets that team's secret names/env keys in the raw response (not rendered in the UI, but present in the network payload).

meAmitPatil (approved):

PR #264, opened 2026-07-06
console: multi-cell support — team directory fan-out, per-cell writes, usw team creation

pavitrabhalla:
The console now supports teams homed in different regions ("cells"), including creating new teams in the usw cell. A cell registry maps region → {Supabase admin client, control-plane API base URL}; team lookups fan out across configured cells, and everything that operates on a team (API key minting, proxy key rows, the API proxy upstream, activity/snapshots/quota/billing queries) targets that team's home cell. Part of multi-region cells this is the console piece that unblocks the west-cell pilot.

## Technical decisions

- **Env-gated dormancy.** The usw cell exists only when `SUPABASE_USWEST_URL` + `SUPABASE_USWEST_SERVICE_ROLE_KEY` are set (`SANDBOX_API_URL_USWEST` optional, defaults to `https://api-usw.superserve.ai`). Absent those, `configuredRegions` is `["use"]`, the region select and the whole Teams settings section render nothing, and every query reduces to today's single-cell path. Env is read per call, not at module load, so enabling a cell is config-only.
- **Fan-out, not a directory service.** At 2 cells, membership lookup is a parallel query per cell merged in region order (default cell first). A central team directory isn't worth it yet; the seam (`team-directory.ts`) is where one would slot in.
- **RBAC chain written in the target cell.** Creating a team in a cell writes profile (upsert — auth is global, profile rows are per-cell), team with `home_region`, the legacy `team_member` row (console lookups still read it), then `team_memberships` (active) before the `team_owner` assignment (`roles` looked up by name; the control-plane trigger enforces that ordering). Without the chain the control plane 403s the team.
- **Key rows land in the team's cell.** Both user-minted keys and `__console_proxy__` rows are written via the registry to the team's home cell, and the browser proxy (`/api/[...path]`, team-management) forwards to that cell's API base URL — the only control plane whose DB holds the key hash. Key format (region prefix from `home_region`, #253) is unchanged.
- **Conservative active-team semantics.** There's still no team switcher; per-action multi-team behavior is preserved (billing throws, quota banner hides, first membership wins elsewhere), now applied across the merged cross-cell membership set. The implicit auto-create path (first login) still creates in the default cell and does not write the RBAC chain — matching today's prod behavior. Follow-ups: a real team switcher for users with teams in multiple cells, and RBAC-chain writes on the auto-create path once the control plane requires it in `use`.
- **Teams settings section is itself gated** behind >1 configured region: the single-cell console has no team-creation surface today, and dormant-until-configured means keeping it that way.

## Testing

- New vitest coverage: registry env gating (absent usw env → only `use`, partial env stays dormant, URL/key wiring), directory merge across a mocked second cell, create-team chain lands entirely in the target cell (and nothing in the default cell), API key insert and proxy key upsert target the team's cell, proxy routes forward to the resolved cell base URL.
- `bun run test` (console): 47 files, 370 tests passing; `bun run typecheck` clean; `bunx oxlint` / `oxfmt` clean on `apps/console`.

### DIFF
--- a/apps/console/src/lib/api/api-keys-actions.ts
+++ b/apps/console/src/lib/api/api-keys-actions.ts
@@ -50,11 +50,17 @@ function hashKey(key: string): string {
 }
 
 /**
- * Ensure a profile row exists for the authenticated user in the default
- * cell. The Go backend schema requires profile(id) to match auth.users(id).
+ * Ensure a profile row exists for the authenticated user in the given cell.
+ * The Go backend schema requires profile(id) to match auth.users(id), and
+ * profile rows are per-cell — a row must exist in whichever cell a write
+ * references it from.
  */
-async function ensureProfile(userId: string, email: string): Promise<void> {
-  const admin = cellFor(DEFAULT_REGION).createAdminClient()
+async function ensureProfile(
+  region: string,
+  userId: string,
+  email: string,
+): Promise<void> {
+  const admin = cellFor(region).createAdminClient()
   const { data: existing } = await admin
     .from("profile")
     .select("id")
@@ -84,7 +90,7 @@ async function getOrCreateTeamForUser(
   email: string,
 ): Promise<TeamMembership> {
   // Ensure profile exists first (FK target for team_member and api_key)
-  await ensureProfile(userId, email)
+  await ensureProfile(DEFAULT_REGION, userId, email)
 
   // Try to find existing team membership in any configured cell
   const memberships = await listTeamMembershipsForUser(userId)
@@ -159,7 +165,11 @@ export async function createApiKeyAction(name: string) {
   const keyPrefix = `${rawKey.slice(0, `ss_live_${region}_`.length + 8)}...`
 
   // The key row must live in the team's cell — that's the database the
-  // team's control plane authenticates against.
+  // team's control plane authenticates against. The creator's profile row
+  // (created_by FK target) must exist in that same cell: today only
+  // createTeamAction guarantees it, and a membership provisioned any other
+  // way (seed, admin tooling, migration) would otherwise FK-violate here.
+  await ensureProfile(team.region, user.id, user.email ?? user.id)
   const admin = cellFor(team.region).createAdminClient()
   const { data, error } = await admin
     .from("api_key")
--- a/apps/console/src/lib/api/api-keys-actions.ts
+++ b/apps/console/src/lib/api/api-keys-actions.ts
@@ -35,8 +35,12 @@ async function getTeamHomeRegion(team: TeamMembership): Promise<string> {
     .select("home_region")
     .eq("id", team.teamId)
     .single()
+  // Fall back to the cell the team was discovered in, not DEFAULT_REGION:
+  // a transient read failure for a usw team must not mint a use-prefixed
+  // key that the edge would route to the wrong cell. team.region is always
+  // a configured region, so it is a safe routing prefix.
   if (error || !data?.home_region || !REGION_CODES.has(data.home_region)) {
-    return DEFAULT_REGION
+    return team.region
   }
   return data.home_region as string
 }
--- a/apps/console/src/lib/api/api-keys-actions.ts
+++ b/apps/console/src/lib/api/api-keys-actions.ts
@@ -2,7 +2,11 @@
 
 import crypto from "node:crypto"
 
-import { createAdminClient } from "@/lib/supabase/admin"
+import {
+  listTeamMembershipsForUser,
+  type TeamMembership,
+} from "@/lib/api/team-directory"
+import { cellFor, DEFAULT_REGION } from "@/lib/cells"
 import { createServerClient } from "@/lib/supabase/server"
 
 // Region codes embedded in new API keys (ss_live_<region>_...). Must stay in
@@ -12,7 +16,6 @@ import { createServerClient } from "@/lib/supabase/server"
 // without a region segment keep working: the control plane hashes the whole
 // string, so the format is opaque to auth.
 const REGION_CODES = new Set(["use", "usw"])
-const DEFAULT_REGION = "use"
 
 function generateRawKey(region: string): string {
   const bytes = crypto.randomBytes(24)
@@ -25,12 +28,12 @@ function generateRawKey(region: string): string {
  * home_region migration hasn't been applied yet or the value is unknown, so
  * key creation never breaks on schema skew.
  */
-async function getTeamHomeRegion(teamId: string): Promise<string> {
-  const admin = createAdminClient()
+async function getTeamHomeRegion(team: TeamMembership): Promise<string> {
+  const admin = cellFor(team.region).createAdminClient()
   const { data, error } = await admin
     .from("team")
     .select("home_region")
-    .eq("id", teamId)
+    .eq("id", team.teamId)
     .single()
   if (error || !data?.home_region || !REGION_CODES.has(data.home_region)) {
     return DEFAULT_REGION
@@ -43,11 +46,11 @@ function hashKey(key: string): string {
 }
 
 /**
- * Ensure a profile row exists for the authenticated user.
- * The Go backend schema requires profile(id) to match auth.users(id).
+ * Ensure a profile row exists for the authenticated user in the default
+ * cell. The Go backend schema requires profile(id) to match auth.users(id).
  */
 async function ensureProfile(userId: string, email: string): Promise<void> {
-  const admin = createAdminClient()
+  const admin = cellFor(DEFAULT_REGION).createAdminClient()
   const { data: existing } = await admin
     .from("profile")
     .select("id")
@@ -68,29 +71,23 @@ async function ensureProfile(userId: string, email: string): Promise<void> {
 }
 
 /**
- * Look up the user's team via team_member. If no team exists,
- * auto-create one (named after their email) and add them as owner.
+ * Look up the user's team across configured cells. If no team exists,
+ * auto-create one (named after their email) in the default cell and add
+ * them as owner.
  */
 async function getOrCreateTeamForUser(
   userId: string,
   email: string,
-): Promise<string> {
-  const admin = createAdminClient()
-
+): Promise<TeamMembership> {
   // Ensure profile exists first (FK target for team_member and api_key)
   await ensureProfile(userId, email)
 
-  // Try to find existing team membership
-  const { data: membership } = await admin
-    .from("team_member")
-    .select("team_id")
-    .eq("profile_id", userId)
-    .limit(1)
-    .single()
+  // Try to find existing team membership in any configured cell
+  const memberships = await listTeamMembershipsForUser(userId)
+  if (memberships[0]) return memberships[0]
 
-  if (membership?.team_id) return membership.team_id
-
-  // No team — create one and add user as owner
+  // No team — create one in the default cell and add user as owner
+  const admin = cellFor(DEFAULT_REGION).createAdminClient()
   const { data: team, error: teamErr } = await admin
     .from("team")
     .insert({ name: email })
@@ -108,7 +105,7 @@ async function getOrCreateTeamForUser(
   if (memberErr)
     throw new Error(`Failed to add team member: ${memberErr.message}`)
 
-  return team.id as string
+  return { teamId: team.id as string, region: DEFAULT_REGION }
 }
 
 export async function listApiKeysAction() {
@@ -118,13 +115,13 @@ export async function listApiKeysAction() {
   } = await supabase.auth.getUser()
   if (!user) throw new Error("Not authenticated")
 
-  const teamId = await getOrCreateTeamForUser(user.id, user.email ?? user.id)
+  const team = await getOrCreateTeamForUser(user.id, user.email ?? user.id)
 
-  const admin = createAdminClient()
+  const admin = cellFor(team.region).createAdminClient()
   const { data, error } = await admin
     .from("api_key")
     .select("id, name, key_hash, created_at, last_used_at")
-    .eq("team_id", teamId)
+    .eq("team_id", team.teamId)
     .is("revoked_at", null)
     .neq("name", "__console_proxy__")
     .order("created_at", { ascending: false })
@@ -147,9 +144,9 @@ export async function createApiKeyAction(name: string) {
   } = await supabase.auth.getUser()
   if (!user) throw new Error("Not authenticated")
 
-  const teamId = await getOrCreateTeamForUser(user.id, user.email ?? user.id)
+  const team = await getOrCreateTeamForUser(user.id, user.email ?? user.id)
 
-  const region = await getTeamHomeRegion(teamId)
+  const region = await getTeamHomeRegion(team)
   const rawKey = generateRawKey(region)
   const keyHash = hashKey(rawKey)
   // ss_live_<region>_ plus the first 8 random chars, e.g. "ss_live_use_AbCdEfGh..."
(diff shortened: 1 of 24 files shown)


apps/console/src/lib/api/team-directory.ts
```diff
@@ -0,0 +1,83 @@
...
+        .eq("profile_id", userId)
+
+      if (error) throw new Error(error.message)
+
+      return (data ?? [])
+        .map((row) => row.team_id)
+        .filter((teamId): teamId is string => typeof teamId === "string")
+        .map((teamId) => ({ teamId, region }))
+    }),
+  )
```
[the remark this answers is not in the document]
pavitrabhalla: Taken in f77f5a9 — both fan-outs now go through a shared acrossCells helper using Promise.allSettled: a failing secondary cell degrades to its absence (logged) while the rest serve. One deliberate asymmetry: the default cell still throws, so single-cell behavior stays byte-identical to the pre-cells console (that's an acceptance criterion on SS-146). Tests cover both directions.

apps/console/src/lib/api/team-directory.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      if (teamErr) throw new Error(teamErr.message)
+
+      return (teams ?? []).map((t) => ({
+        id: t.id as string,
+        name: t.name as string,
+        region,
+      }))
+    }),
+  )
```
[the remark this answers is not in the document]
pavitrabhalla: Same fix — listTeamsForUser shares the acrossCells helper from the sibling thread (f77f5a9), with a test asserting a usw failure leaves the use directory intact.

apps/console/src/lib/cells.ts
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    region: "usw",
+    apiBaseUrl:
+      process.env.SANDBOX_API_URL_USWEST ?? "https://api-usw.superserve.ai",
+    createAdminClient: () =>
+      createClient(url, serviceRoleKey, {
+        auth: { autoRefreshToken: false, persistSession: false },
+      }),
+  }
+}
```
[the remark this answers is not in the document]
pavitrabhalla: Taken in f77f5a9 — clients are now cached per credential set (url+key map, not a singleton), so calls reuse one client while env swaps in tests and future key rotations still get a matching client.

pavitrabhalla:
All four review findings addressed in 57f6a26:

**High (stale authz cache)** — taken. Both proxy-auth caches (team membership and the key-ensure marker, the latter pre-existing) now expire on a 60s TTL, so a removed or moved member loses proxy access within a bounded window instead of at process recycle. Cost is one membership read per user per minute.

**Medium (non-atomic create-team)** — taken as compensating cleanup: a mid-chain failure unwinds in reverse dependency order (assignments → memberships → team_member → team), with unwind failures logged and the team id surfaced in the error for manual repair. Agreed the durable fix is a per-cell transactional RPC — noted on SS-146 as a follow-up rather than expanding this PR into a control-plane migration.

**Medium (wrong-region key fallback)** — taken exactly as recommended: getTeamHomeRegion now falls back to the cell the team was discovered in (always a configured region, so always a safe routing prefix); DEFAULT_REGION no longer masks read failures for non-default teams.

**Medium (fan-out scaling)** — agreed and deliberately deferred: 2-cell fan-out is the pilot design, and SS-146 already carries the revisit note (per-user home-cell index / primary directory) before region N+1. The metrics you list (per-cell lookup latency, secondary-cell failures, selected/proxy target region) are folding into SS-150's per-cell observability work.

Rollout guidance matches the plan on the ticket: allowlist stays, first-membership-wins is documented until the switcher (SS-162) ships.

awille-ai (commented):
Architecture review focused on multi-region/team-home-cell behavior, scaling, security, and race conditions.

Summary: I do not see any Critical findings, so I would not block this PR. I do see one High finding, so I also would not approve it as-is.

What looks good:
- Region config is dormant unless the west Supabase env is fully present.
- Creatable regions are enforced server-side, not only by hiding UI.
- Team-scoped reads and writes mostly target `cellFor(team.region)`, which is the right direction for tenant data locality.
- Authenticated proxy requests route to the resolved home-cell API, and client-supplied sensitive headers are stripped.
- The new tests cover env gating, fan-out, per-cell key writes, proxy routing, and target-cell team creation.

Key risks:

High: stale in-process team authorization cache can preserve access after membership or region changes.

`proxy-auth.ts` caches `{ teamId, region }` by `userId` in `teamCache` with no TTL or invalidation. `ensuredUsers` also suppresses re-checking the proxy key row after the first successful ensure. If a user is removed from a team, moved between teams, or their active/home region changes, a warm console process can continue routing and issuing the deterministic console proxy key for the old team until process restart. In a multi-tenant console, stale authorization like this is a tenant-safety risk.

Recommendation: remove the unbounded cache on privileged proxy paths, add a short TTL, or key it off a durable membership/version marker if one exists. At minimum, membership should be revalidated frequently enough that team removal is not dependent on process recycle.

Medium: create-team writes are not atomic and can leave partial RBAC state.

`createTeamAction` writes `profile`, `team`, `team_member`, `team_memberships`, role lookup, and `user_role_assignments` as separate calls. Any failure after the `team` insert can leave an orphaned or partially authorized team. This is especially risky because legacy console lookups read `team_member`, while the control plane depends on the newer RBAC chain.

Recommendation: move this into a per-cell Postgres RPC/transaction that creates the team and RBAC chain atomically. If that is too much for this PR, add cleanup/retry semantics and a repair runbook before rollout.

Medium: west API keys can be minted with the wrong routing prefix on `home_region` read failure.

`getTeamHomeRegion` falls back to `DEFAULT_REGION` when the `home_region` read fails or returns an unknown value. For a west team, a transient read failure could mint an `ss_live_use_...` key while storing the key hash in the west cell. If the API edge routes by key region segment, that key would route east and fail.

Recommendation: prefer `team.region` as the fallback for teams discovered in configured cells, and reserve `DEFAULT_REGION` only for true legacy/unknown cases.

Medium: fan-out is acceptable at 2 cells but will not scale linearly.

The team directory does parallel per-cell membership lookup. That is fine for the west-cell pilot, but every new region adds database calls to common console paths. Before expanding beyond a small number of regions, add a primary-region team directory or user-to-home-cell index.

Rollout guidance:
- Keep the allowlist rollout.
- Add metrics for per-cell lookup latency, secondary-cell lookup failures, selected team region, and proxy target region.
- Add an operator repair path for partial team creation before customer-facing rollout.
- Document that users with multiple teams still get first-membership-wins behavior until the team switcher ships.

Recommendation: proceed with changes after fixing the stale authorization cache. I would not block as Critical, but I would not approve this PR until the High issue is addressed.

apps/console/src/lib/api/billing-actions.ts
```diff
@@ -47,27 +53,15 @@ export interface BillingSettingsResponse {
...
-  const teamIds = [
-    ...new Set(
-      data
-        .map((row) => row.team_id)
-        .filter((teamId): teamId is string => typeof teamId === "string"),
-    ),
-  ]
-  if (teamIds.length !== 1) {
+  const uniqueTeams = new Map(memberships.map((m) => [m.teamId, m]))
+  if (uniqueTeams.size !== 1) {
```
pavitrabhalla, line 73: A multi-cell user (teams in `use` + `usw`) hits `size !== 1` and gets "Select a team". But if the `usw` cell is momentarily unreachable, `acrossCells` drops that membership, `size` becomes 1, and this silently returns the `use` team's billing — so the guard's outcome flips on transient secondary-cell availability rather than on the actual membership set. Worth having `acrossCells` surface partial failure so this can refuse instead of guessing?

apps/console/src/lib/api/quota-actions.ts
```diff
@@ -9,29 +13,17 @@ export interface QuotaUsageResponse {
...
-      data
-        .map((row) => row.team_id)
-        .filter((teamId): teamId is string => typeof teamId === "string"),
-    ),
-  ]
+  const uniqueTeams = new Map(memberships.map((m) => [m.teamId, m]))
   // Ambient poller with no team-selector: an ambiguous (multi-team) membership
   // renders nothing rather than throwing on every poll (which would spam logs).
-  if (teamIds.length !== 1) {
+  if (uniqueTeams.size !== 1) {
```
pavitrabhalla, line 29: Same secondary-cell-outage edge as billing `getTeam`: a genuinely multi-team user renders nothing here, but if `usw` is down the dropped membership makes `size === 1` and the banner shows the `use` team's quota. Fine for an ambient poller, or should a partial fan-out failure also render nothing so the number is never conditioned on a sibling cell being up?

apps/console/src/lib/api/api-keys-actions.ts
```diff
@@ -43,11 +50,11 @@ function hashKey(key: string): string {
...
 
 /**
- * Ensure a profile row exists for the authenticated user.
- * The Go backend schema requires profile(id) to match auth.users(id).
+ * Ensure a profile row exists for the authenticated user in the default
+ * cell. The Go backend schema requires profile(id) to match auth.users(id).
  */
 async function ensureProfile(userId: string, email: string): Promise<void> {
-  const admin = createAdminClient()
+  const admin = cellFor(DEFAULT_REGION).createAdminClient()
```
pavitrabhalla: `ensureProfile` writes to the default cell, but `createApiKeyAction` inserts the `api_key` row (with `created_by` FK -> `profile`) into `team.region`'s cell. That holds today only because `createTeamAction` is the sole path that creates a `usw` team and it upserts the profile there first. If a `usw` membership ever arrives another way (backend-provisioned team, seed, admin tooling), the key insert would FK-violate. Should this `ensureProfile` run in `team.region` at the write site rather than always the default cell?

apps/console/src/lib/api/billing-actions.ts
pavitrabhalla: Fixed in 67216f4 — listTeamMembershipsForUserDetailed now returns degradedRegions alongside memberships, and billing's getTeam fails closed with a retry message whenever any cell was dropped, so 'exactly one membership' is only ever trusted on a complete read.

apps/console/src/lib/api/quota-actions.ts
pavitrabhalla: Fixed in 67216f4 — the quota banner renders nothing on a degraded read (same detailed lookup as billing). For an ambient poller, silence beats a possibly-wrong number; SS-162's switcher will make degraded cells visible explicitly.

apps/console/src/lib/api/api-keys-actions.ts
pavitrabhalla: Fixed in 67216f4 — ensureProfile is parameterized by cell and createApiKeyAction now ensures the creator's profile in team.region right before the key insert, so key creation no longer assumes createTeamAction was the provisioning path (relevant for the SS-163 migration tooling too, which is exactly a 'membership arrives another way' case). The cell-targeting test now asserts the profile check happens in the usw cell.

awille-ai (commented):
Follow-up architecture/security review after the latest fixes, cross-checked against the merged sandbox user-management/RBAC PRs (#140, #143, #145, #147).

Review state: no Critical findings, so I am not blocking. I did find one High, so I would not approve as-is under the stated bar.

What changed since the previous pass:
- The unbounded proxy auth caches now have a 60s TTL. That bounds the stale-access window and removes the previous High as written.
- `getTeamHomeRegion` now falls back to `team.region` instead of `DEFAULT_REGION`, so west keys should not be minted with an east routing prefix on a read blip.
- Billing/quota now fail closed or render nothing on degraded multi-cell membership reads.
- API-key creation now ensures the creator profile exists in the target team cell before inserting `api_key`, which matches the per-cell profile model.

High: console authorization still treats legacy `team_member` as the membership source of truth, while sandbox RBAC/user-management now uses active `team_memberships` + `user_role_assignments`.

The merged sandbox RBAC service checks `team_memberships.status = 'active'` and active `user_role_assignments` for permissions. But this console PR's cross-cell directory lookup still reads only `team_member`, and many console server actions then use a service-role client directly against team-scoped tables (`api_key`, billing usage, activity, snapshots, quota) based on that legacy row. A user deactivated through the sandbox user-management endpoints will have their `team_memberships` row marked inactive and role grants revoked, but the legacy `team_member` row is not the source of truth for deactivation. That means the sandbox API may correctly deny the user, while console server actions can still list/read/mutate team data because they bypass backend RBAC and trust `team_member`.

Examples of likely impact:
- A deactivated user can still be selected by `listTeamMembershipsForUser*` and read activity/snapshots/billing/quota through console server actions.
- `listApiKeysAction` / `createApiKeyAction` / `revokeApiKeyAction` authorize through the same legacy membership lookup and then mutate `api_key` with service-role access.
- The proxy path is less exposed because the sandbox API enforces RBAC on protected routes, but the console's direct Supabase server actions do not get that protection.

Recommendation: for any console path that authorizes user access to team-scoped data, read active `team_memberships` (and, where mutations are involved, require the relevant RBAC permission) in the target cell instead of relying on `team_member`. If keeping `team_member` temporarily is necessary for legacy discovery, it should be treated as a compatibility hint only, not as authorization. At minimum, join/filter against `team_memberships.status = 'active'` in `team-directory.ts`, and do not allow service-role reads/writes for users who no longer have active membership.

Medium: `/api/team-management` still adds a console-side platform/staff gate before forwarding to a customer RBAC route.

Sandbox PR #147 describes `/teams/:team_id/management` as customer-facing and derives actor/team from API-key context, with backend RBAC authoritative. The console wrapper still checks `canViewOtherUsersAccount(user)`, which is a Supabase app_metadata/staff-domain/platform permission gate, before forwarding to the sandbox route. If this endpoint is intended for normal team owners/admins, that will block legitimate customer RBAC users and effectively add a second auth scheme in front of the backend model. If the endpoint is intentionally staff-only in console for now, please document that clearly so nobody expects the customer RBAC capabilities from phase 4 to be exposed here yet.

I do not see evidence that this PR invents a new backend authentication scheme: it continues to use Supabase session in the console, server-derived API keys for the proxy, and sandbox API-key auth/RBAC behind the target cell. The issue is that some console server actions still bypass the sandbox RBAC layer and authorize from the legacy membership table.

pavitrabhalla:
Follow-up review addressed:

**High (legacy team_member as authz source)** — taken, in 28f65ca. The directory lookup (both membership and team listing, one choke point) is now RBAC-authoritative: wherever a `team_memberships` row exists, only `status='active'` passes — so backend deactivation now revokes console access too. Legacy `team_member` counts only when a member has no RBAC row of any status, which prod data says is exactly 10 pre-RBAC members across 10 teams; treating legacy as discovery-only means it can never resurrect revoked access, and those 10 keep working until a backfill migration (follow-up noted below). Tests cover deactivated-with-legacy-row (denied), pure-legacy (allowed), and RBAC-only-active (allowed).

**Medium (/api/team-management staff gate)** — documented as intentional for now rather than changed: the console wrapper predates customer-facing RBAC and is deliberately staff-scoped until the console grows real team-management UI. Filed as a follow-up (with the team_member backfill migration) so customer RBAC exposure is a decision, not drift.

awille-ai (approved):
Follow-up review after 28f65ca.

Review state: Approved. I do not see any remaining Critical or High findings.

The prior High around legacy `team_member` as the console authorization source looks materially addressed. `team-directory.ts` now uses `team_memberships` as authoritative wherever an RBAC row exists, only allowing `status = 'active'`; legacy `team_member` is now a compatibility fallback only when no RBAC row exists for that team. The added tests cover the important security case: inactive RBAC membership plus legacy row is denied.

Auth/user-management check:
- This does not appear to introduce a new backend authentication scheme. The console still uses Supabase session auth, derives/injects the existing console proxy API key server-side, and relies on the sandbox API key/RBAC path for proxied backend requests.
- The console direct service-role reads now share the updated membership directory, so backend deactivation should revoke console access to those team-scoped views as well.
- Per-cell profile/key placement and wrong-region key fallback from the earlier review are addressed.

Residual medium / rollout notes, not approval blockers:
- The pure-legacy fallback should stay temporary. If someone manually deletes a `team_memberships` row while leaving `team_member`, the fallback could treat that as a legacy-only membership; the normal backend deactivation path updates to inactive, so this is not the customer-path issue we flagged, but the follow-up backfill/removal is still important.
- `/api/team-management` remains intentionally staff/platform-gated in the console while the sandbox route is customer-RBAC-capable. That is fine if documented as a deliberate exposure choice, but should be revisited when customer team-management UI ships.
- The create-team flow still uses best-effort unwind rather than a transactional RPC. Acceptable for the pilot, but I would still move this into a per-cell RPC before broad customer rollout.

PR #265, opened 2026-07-07
sdk: auto_delete_seconds on create, patchable auto-delete and timeout on update

meAmitPatil:
SDK, console, MCP, and docs support for paused-sandbox auto-delete, plus a patchable auto-pause timeout. Mirrors the control-plane API (superserve-ai/sandbox#178).

- **TS + Python SDKs**: `autoDeleteSeconds`/`auto_delete_seconds` on create; `update()` sets or clears (`null`/`None`) both `autoDeleteSeconds` and `timeoutSeconds` (absent = unchanged); `autoDeleteSeconds` + `autoDeleteAt` on `SandboxInfo`.
- **Console**: auto-delete field in the create dialog and a deadline cell on the detail grid.
- **MCP**: `sandbox_create`/`sandbox_update` accept the fields; `sandbox_info` reports them.
- **Docs**: Auto-delete section on the lifecycle page, option rows, `update()` reference, SKILL.md.

Semantics match the API: pause arms the window, resume cancels it, `0` = delete on pause, arming on an already-paused sandbox counts from now.

Tests: vitest (SDK 215, console 351), pytest (287), MCP (88); typecheck + lint clean.

### DIFF
--- a/apps/console/src/components/sandboxes/create-sandbox-dialog.tsx
+++ b/apps/console/src/components/sandboxes/create-sandbox-dialog.tsx
@@ -19,6 +19,7 @@ import {
   Field,
   HighlightedCode,
   Input,
+  useToast,
 } from "@superserve/ui"
 import { AnimatePresence, LayoutGroup, motion } from "motion/react"
 import { usePostHog } from "posthog-js/react"
@@ -241,6 +242,7 @@ function CopyButton({ text }: { text: string }) {
 interface FormState {
   name: string
   timeout: string
+  autoDelete: string
   allowRules: string[]
   denyRules: string[]
   secretEntries: SecretBindingEntry[]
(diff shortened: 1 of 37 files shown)


awille-ai (commented):
Architecture/security review notes. I found one high-severity MCP safety concern and one medium validation edge case. I am not approving because of the high finding, but I am leaving this as comments rather than a blocking review per request.

packages/mcp/src/tools/lifecycle.ts
```diff
@@ -554,6 +581,29 @@ export function registerLifecycleTools(
...
+              "auto-delete. Omit to leave unchanged.",
+          ),
+        timeout_seconds: z
+          .number()
+          .int()
+          .positive()
+          .max(MAX_TIMEOUT_SECONDS)
+          .nullable()
+          .optional()
+          .describe(
```
awille-ai, line 603: High: with `auto_delete_seconds` this tool is no longer safely non-destructive or idempotent. A caller can pass `auto_delete_seconds: 0` and delete an already-paused sandbox immediately, or combine `timeout_seconds` + `auto_delete_seconds` to schedule deletion of state later. Also, on an already-paused sandbox the docs say setting a numeric value starts the countdown "now", so retrying the same request can move `auto_delete_at`, which makes `idempotentHint: true` misleading. MCP clients may use these annotations to decide approval/retry behavior, so this can become a real data-loss/safety footgun. Please mark this destructive when auto-delete is supported, and consider dropping the idempotent hint or splitting destructive lifecycle cleanup into a separate tool.

apps/console/src/components/sandboxes/create-sandbox-dialog.tsx
```diff
@@ -297,6 +298,9 @@ export function buildCreateSandboxRequest(
     name: state.name.trim(),
```
awille-ai, line 345: Medium: `type="number"` min/max does not protect this code path because creation is triggered from `handleCreate`, not a native validated form submit. Users can still send out-of-range, fractional, or non-finite values here; for example `Number("1e309")` becomes `Infinity`, which `JSON.stringify` serializes as `null`. Since auto-delete is a billing/data-retention control, please validate `Number.isSafeInteger`, bounds (`0..2592000`), and finite values before building the request or before mutating.

pavitrabhalla (approved):
Verified the earlier high/medium concerns from awille-ai's review are addressed in the current commit: `sandbox_update` now correctly marks `destructiveHint: true` / `idempotentHint: false`, and the console auto-delete field validates via `Number.isSafeInteger` + bounds before building the request. Ran an independent review across the SDK/MCP/console diff — nothing blocking. Branch is behind main, worth updating before merge.

awille-ai (approved):
Security/scalability review pass complete. I did not find any critical or high-severity issues, so I’m approving.

The important safety fixes are in place on the current head: MCP update no longer activates a paused sandbox just to patch it, MCP marks the update tool destructive/non-idempotent, console input validates finite integer bounds before sending, and CI is green.

One non-blocking hardening suggestion for a follow-up: the console and MCP validate lifecycle windows before serialization, but the raw TS SDK currently forwards arbitrary numeric values. Because `JSON.stringify` serializes `NaN`/`Infinity` as `null`, validating `Number.isSafeInteger` and the documented bounds in the SDK as well would prevent a malformed computed value from accidentally clearing `autoDeleteSeconds` or `timeoutSeconds`. I don't consider that a merge blocker because the control plane still validates normal API inputs.

awille-ai (approved):
Security/scalability review pass complete. I did not find any critical or high-severity issues, so I’m approving.

The important safety fixes are in place on the current head: MCP update no longer activates a paused sandbox just to patch it, MCP marks the update tool destructive/non-idempotent, console input validates finite integer bounds before sending, and CI is green.

One non-blocking hardening suggestion for a follow-up: the console and MCP validate lifecycle windows before serialization, but the raw TS SDK currently forwards arbitrary numeric values. Because `JSON.stringify` serializes `NaN`/`Infinity` as `null`, validating `Number.isSafeInteger` and the documented bounds in the SDK as well would prevent a malformed computed value from accidentally clearing `autoDeleteSeconds` or `timeoutSeconds`. I don't consider that a merge blocker because the control plane still validates normal API inputs.

PR #270, opened 2026-07-08
feat: server-side pagination for audit logs page

mohamedsobhi777:
## Summary

Migrates the Audit Logs page from client-side filtering of a capped ≤100-row fetch to true server-side pagination against the new `GET /activity` endpoint, mirroring the Sandboxes list page.

Previously the page fetched up to 100 activity rows via a Supabase server action (`listActivityAction`) and did all category/search/date filtering client-side — so it rendered a long unpaginated list and never loaded history beyond the latest 100. Now:

- `useActivityPage` → `listActivityPaged` → the API proxy → `GET /activity`, reusing the shared `useListParams` + `Pagination` machinery. The team-wide Supabase read is removed (the per-sandbox `listActivityBySandboxAction` stays).
- Filters move to the server and into the URL (shareable, survive refresh): category tabs → `category` (Errors tab → `status=error`), date range → `start`/`end`, search → `q`. The global total drives the pager's "X of Y".
- Per-tab counts dropped (tabs are plain filters now); `ActivityResponse` gains `template_id` + `actor_id` to match the endpoint.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `oxlint` + `oxfmt --check` clean (pre-commit hook)
- [x] Full console suite: **397 tests pass**, including new `activity.test.ts` (query-builder param mapping + `X-Total-Count` parsing / fallback)
- [ ] Manual: page through, filter by tab/search/date, confirm URL state + "X of Y" (after the backend endpoint is deployed)

## Depends on

- superserve-ai/sandbox#189 (the `GET /activity` endpoint). **Merge + deploy that first** — until the route is live, this page will 404 its data.

## Follow-up (not in this PR)

The API-reference **Activity** nav entry (`docs.json`) is intentionally deferred: `check-docs-nav` validates the nav against the backend's `openapi.yaml` on `main`, so a `GET /activity` nav entry can only land once #189 is merged. It'll go in a small docs-only PR right after — merge it promptly, since once #189 is on `main` the check will expect the nav entry on subsequent console PRs.

## Review notes

Independently reviewed (hooks order / Suspense boundary / empty-state / dead-code all confirmed clean). One robustness fix applied: `parseDateRange` now rejects an inverted (`end < start`) URL range, degrading to "no filter" instead of a confusing empty result.

### DIFF
--- a/apps/console/src/app/(dashboard)/audit-logs/page.tsx
+++ b/apps/console/src/app/(dashboard)/audit-logs/page.tsx
@@ -1,21 +1,36 @@
 "use client"
 
+import { Suspense } from "react"
+
+import { TableSkeleton } from "@/components/table-skeleton"
+
+export default function AuditLogsPage() {
+  return (
+    <Suspense fallback={<TableSkeleton columns={6} tabs={4} />}>
+      <AuditLogsPageContent />
+    </Suspense>
+  )
+}
+
 import { ClipboardTextIcon } from "@phosphor-icons/react"
-import { Table, TableHead, TableHeader, TableRow } from "@superserve/ui"
-import { useMemo, useState } from "react"
+import { cn, Table, TableHead, TableHeader, TableRow } from "@superserve/ui"
+import { useSearchParams } from "next/navigation"
+import { useEffect, useState } from "react"
 
 import {
   ActivityDetailRow,
   ActivitySummaryRow,
 } from "@/components/audit/activity-row"
-import { DateRangeFilter } from "@/components/date-range-filter"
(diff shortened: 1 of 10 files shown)


awille-ai (approved):
Security/scalability review pass complete. I did not find any critical or high-severity issues, so I’m approving.

The console migration follows the backend contract cleanly: server-side filtering/pagination, URL-backed state, narrow proxy allowlist expansion, placeholder data to avoid flashing, and explicit dependency on the backend rollout. CI is green.

The remaining open suggestion about snapping back to page 1 when filters produce zero results is a reasonable UX improvement, but I don't consider it a correctness, security, or scalability blocker.

PR #273, opened 2026-07-09
console: provision full RBAC chain on first-login team auto-create

meAmitPatil:
## Problem

A brand-new user (e.g. Google OAuth first login) lands on `/sandboxes` and immediately hits **"Something went wrong — You do not have permission to perform this action."** The auth callback provisions nothing, so the user has zero memberships; the first server-side request is what creates their team.

Every auto-provision path wrote only `team` (without `home_region`) + the legacy `team_member` row — **not** the RBAC chain (`team_memberships` + `user_role_assignments`) the control plane authorizes against. The result is exactly what `createTeamAction`'s own comment warns about: *"a team the console can list but the control plane rejects (no RBAC chain)"* → 403 on the user's first request.

This isn't new to the team-switcher PR — the incomplete auto-provision predates it. It surfaces now that RBAC `team_memberships` is the console's authorization boundary, so legacy-only teams get rejected.

## Fix

The root cause was **three divergent team-creation code paths, only one correct**. Extracted the full, correct chain into a single `provisionTeam()` helper (`team-provisioning.ts`) and routed all three call sites through it:

- `createTeamAction` (`teams-actions.ts`) — behavior unchanged (same full chain + transactional unwind), just deduplicated (~120 lines removed).
- `getTeamForUser`'s first-login fallback (`proxy-auth.ts`) — the path the sandboxes 403 came from. Now writes the full RBAC chain + `home_region`, and invalidates the membership directory cache so other surfaces see the new team immediately.
- `getOrCreateTeamForUser` (`api-keys-actions.ts`) — the same legacy-only bug on the API-keys surface. Now delegates to the helper.

`provisionTeam` is now the only way to create a team, so a new user can't end up with a legacy-only team on any surface.

## Tests

- New `team-provisioning.test.ts`: full-chain write into the target cell, and the previously-untested mid-chain-failure **unwind in reverse dependency order**.
- `proxy-auth.test.ts`: new-user path asserts `getTeamForUser` routes a memberless user through `provisionTeam`.
- Full console suite green (428 tests), typecheck + oxlint clean.

## Not in scope

Users who already got a legacy-only team from the old path will keep 403-ing until their missing RBAC rows are backfilled — a one-time data fix, separate from this code change.

> Note: the failing `docs:check-nav` CI job is unrelated and pre-existing on `main` (OpenAPI `GET /activity` missing from docs nav).

### DIFF
--- a/apps/console/src/lib/api/team-provisioning.ts
+++ b/apps/console/src/lib/api/team-provisioning.ts
@@ -1,0 +1,134 @@
+import { cellFor } from "@/lib/cells"
+
+// Role granted to a team's creator. Seeded by the control-plane RBAC
+// migration in every cell; looked up by name so the id can differ per cell.
+const TEAM_OWNER_ROLE = "team_owner"
+
+export interface ProvisionedTeam {
+  id: string
+  name: string
+  region: string
+}
+
+/**
+ * Create a team homed in `region` and write everything the control plane
+ * needs to authorize `userId` as its owner, all in that cell: the profile
+ * row (auth is global but profile rows are per-cell), the team with
+ * home_region, the legacy team_member row the console's own lookups read,
+ * and the RBAC chain (active membership + team_owner assignment). Without the
+ * RBAC chain the console can list the team but the control plane rejects
+ * every request for it.
+ *
+ * This is the ONLY correct way to create a team: both the explicit
+ * create-team action and the first-login auto-provision route through it, so
+ * a new user can never end up with a legacy-only team the control plane 403s.
+ *
+ * Postgrest calls aren't a transaction, so a failure mid-chain is unwound in
+ * reverse dependency order — a partial team the console lists but the control
+ * plane rejects is worse than no team. A per-cell RPC doing the whole chain
+ * in one transaction is the durable replacement.
+ *
+ * ponytail: best-effort compensating unwind, not a real transaction —
+ * replace with a per-cell RPC if half-written teams start showing up.
+ *
+ * The caller owns authorization (who may create where) and cache
+ * invalidation; this only writes the rows.
+ */
+export async function provisionTeam(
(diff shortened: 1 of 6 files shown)


apps/console/src/lib/api/team-provisioning.ts
```diff
@@ -0,0 +1,134 @@
...
+  const { error: profileErr } = await admin
+    .from("profile")
+    .upsert({ id: userId, email }, { onConflict: "id", ignoreDuplicates: true })
+  if (profileErr) {
+    throw new Error(`Failed to create profile: ${profileErr.message}`)
+  }
+
+  const { data: team, error: teamErr } = await admin
+    .from("team")
+    .insert({ name, home_region: region })
```
awille-ai, line 56: Non-blocking edge case: this fixes the missing RBAC chain, but first-login provisioning is still not idempotent at the user/team level. If two server requests for the same brand-new user both observe no memberships before either `provisionTeam()` call completes, both can insert a new `team` and then give the user `team_owner` on two separate teams. The membership directory caches the in-flight empty lookup, so parallel requests after a cold login can all share the same empty result and then race into this path. This is probably not a security issue, but it can create duplicate owner teams and make the deterministic fallback choose one arbitrarily. Longer-term, this should be a transactional RPC keyed by user/account bootstrap, or otherwise guarded by a unique/idempotency constraint so concurrent first-login calls converge on the same team.

awille-ai (commented):
Reviewed for the first-login RBAC issue plus security/scaling/edge cases. This PR does address the missing permission-chain problem: the new helper writes `team_memberships` and a `team_owner` `user_role_assignments` row in addition to the legacy `team_member` row. I did not find any critical findings, so I am not blocking. I left one non-blocking concurrency edge-case comment below.

PR #275, opened 2026-07-09
Add usw region routing to the SDKs and console data plane

meAmitPatil:
Makes west-cell (`usw`) teams work from all three clients: SDKs, console, and MCP. Before this, a west team's key/console/MCP calls landed on the primary (us-east) cell, so their sandboxes were unreachable.

**SDKs** — add `usw` → `{api-usw.superserve.ai, usw-sandbox.superserve.ai}` to the known-regions map (TS + Python). A `usw`-prefixed key now self-routes to the west cell with no config; the endpoints were verified serving end-to-end (control plane, data plane, and the wrong-region 401). Legacy keys and pinned `base_url` are unchanged. Also exports `resolveConfig` so other packages can reuse the region resolution.

**Console** — data-plane URLs (terminal, files, preview) previously hardcoded `sandbox.superserve.ai`, so west sandboxes were unreachable from the UI. New `sandboxHostFor(sandboxId)` derives the host from the region in the public id (`sb-usw-…`). The default cell and legacy ids keep honoring `NEXT_PUBLIC_SANDBOX_HOST` (staging unaffected).

**MCP** — preview URLs and the network-log call went through the raw `config.baseUrl` and defaulted to the primary cell, so a `usw` key hit central. They now use the SDK's region-resolved endpoints, and MCP's duplicate host-derivation is deleted (single source of truth in the SDK).

**Review follow-ups in this PR:** bump `@superserve/sdk` to 0.8.0 and `MIN_SDK_VERSION` together (guards the new `resolveConfig` export so an MCP release can't ship against an older published SDK); normalize an empty/whitespace `SUPERSERVE_BASE_URL` to unset (was becoming a broken empty override); derive the explicit-base-URL → sandbox-host mapping from the region map so a new cell needs only its map entry; and align MCP's preview-port floor to the proxy's 1024 minimum.

Tests: SDK, Python, MCP, and console suites green; typecheck clean.

### DIFF
--- a/apps/console/src/lib/sandbox-host.ts
+++ b/apps/console/src/lib/sandbox-host.ts
@@ -17,9 +17,9 @@
  * `use-sandbox` form; add it here once it does.
  */
 
-const REGION_SANDBOX_HOSTS: Record<string, string> = {
-  usw: "usw-sandbox.superserve.ai",
-}
+const REGION_SANDBOX_HOSTS = new Map<string, string>([
+  ["usw", "usw-sandbox.superserve.ai"],
+])
 
 const DEFAULT_SANDBOX_HOST =
   process.env.NEXT_PUBLIC_SANDBOX_HOST ?? "sandbox.superserve.ai"
@@ -32,5 +32,8 @@ export function regionFromSandboxId(sandboxId: string): string | undefined {
 /** The data-plane host to reach a sandbox, derived from its id's region. */
 export function sandboxHostFor(sandboxId: string): string {
   const region = regionFromSandboxId(sandboxId)
-  return (region && REGION_SANDBOX_HOSTS[region]) ?? DEFAULT_SANDBOX_HOST
+  return (
+    (region ? REGION_SANDBOX_HOSTS.get(region) : undefined) ??
+    DEFAULT_SANDBOX_HOST
+  )
 }
--- a/apps/console/src/lib/sandbox-host.ts
+++ b/apps/console/src/lib/sandbox-host.ts
@@ -1,0 +1,36 @@
+/**
+ * Data-plane host resolution for a sandbox, keyed off the region token in its
+ * public id (`sb-<region>-<uuid>`).
+ *
+ * Public sandbox ids carry their home cell (`sb-usw-…`), so every data-plane
+ * URL (files, terminal, preview) can resolve the right proxy host from the id
+ * alone — no team/region prop-threading, and it works in pure modules that
+ * can't call hooks.
+ *
+ * The default cell (`use`, and legacy un-prefixed ids) keeps honoring
+ * `NEXT_PUBLIC_SANDBOX_HOST` so staging/dev can point at `staging-sandbox…`.
+ * Non-default cells resolve to their own host and ignore that override — a
+ * `usw` sandbox is never reachable via the `use` proxy.
+ *
+ * Mirrors the SDK's region map (packages/sdk/src/config.ts). `use` stays on
+ * the legacy `sandbox.superserve.ai` host until its proxy also accepts the
+ * `use-sandbox` form; add it here once it does.
+ */
+
+const REGION_SANDBOX_HOSTS: Record<string, string> = {
+  usw: "usw-sandbox.superserve.ai",
+}
+
+const DEFAULT_SANDBOX_HOST =
+  process.env.NEXT_PUBLIC_SANDBOX_HOST ?? "sandbox.superserve.ai"
+
+/** Region token from a public sandbox id (`sb-<region>-<uuid>`), else undefined. */
+export function regionFromSandboxId(sandboxId: string): string | undefined {
+  return /^sb-([a-z0-9]{1,17})-/.exec(sandboxId)?.[1]
+}
+
+/** The data-plane host to reach a sandbox, derived from its id's region. */
+export function sandboxHostFor(sandboxId: string): string {
+  const region = regionFromSandboxId(sandboxId)
+  return (region && REGION_SANDBOX_HOSTS[region]) ?? DEFAULT_SANDBOX_HOST
+}
--- a/packages/mcp/src/client.ts
+++ b/packages/mcp/src/client.ts
@@ -256,6 +256,8 @@ export function createSdkClient(config: ClientConfig): SandboxClient {
   // Region-resolved endpoints (key prefix → cell), the single source the SDK
   // uses internally — so our direct preview/network calls hit the same cell
   // as the SDK-backed sandbox ops instead of defaulting to the primary.
+  // resolveConfig is exported as of @superserve/sdk 0.8.0 — the current
+  // MIN_SDK_VERSION floor in mcp-publish.yml (max across features).
   const resolved = resolveConfig({
     apiKey: config.apiKey,
     baseUrl: config.baseUrl,
--- a/packages/mcp/src/client.ts
+++ b/packages/mcp/src/client.ts
@@ -16,6 +16,7 @@
 import {
   AuthenticationError,
   NotFoundError,
+  resolveConfig,
   Sandbox,
   SandboxError,
   Secret,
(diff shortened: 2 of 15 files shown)


packages/mcp/src/client.ts
```diff
@@ -16,6 +16,7 @@
 import {
   AuthenticationError,
   NotFoundError,
+  resolveConfig,
```
pavitrabhalla, line 19: `resolveConfig` is exported from the SDK for the first time in this PR, but `packages/sdk/package.json` is still 0.7.8 and `mcp-publish.yml` keeps `MIN_SDK_VERSION=0.7.8` while pinning the npm-published SDK (tsup externalizes it). An MCP release cut before a new SDK publish passes the gate, installs npm 0.7.8, and crashes at startup with `resolveConfig is not a function`. Bump the SDK version and `MIN_SDK_VERSION` together in this PR?

packages/sdk/src/config.ts
```diff
@@ -18,8 +18,8 @@ const DEFAULT_SANDBOX_HOST = "sandbox.superserve.ai"
  * API — DNS resolving is not the bar: superserve.ai has a catch-all
  * wildcard, so every derived hostname resolves and answers with a
  * valid-TLS 404 from the wrong infrastructure. Verify with a live /health
- * check, not dig. `usw` will be added once `https://api-usw.superserve.ai`
- * / `usw-sandbox.superserve.ai` serve their cell.
+ * check, not dig. `usw` was added once `https://api-usw.superserve.ai`
```
pavitrabhalla, line 21: usw is added to the region map but not to `SUPPORTED_SHARED_HOSTS` (same in the Python SDK's `_SUPPORTED_SHARED_HOSTS`), so server-side `dataPlaneTarget` for usw sandboxes always takes the per-sandbox-subdomain path — fresh DNS+TLS per sandbox, no shared-origin pooling, unlike use/staging. Is that intentional (usw proxy not doing shared-host routing yet) or an oversight? If intentional, worth a comment on the set; if not, it's one entry per language.

packages/mcp/src/client.ts
```diff
@@ -252,6 +253,13 @@ function toSecretSummary(s: SecretInfo): SecretSummary {
 /** Real client backed by `@superserve/sdk`. */
 export function createSdkClient(config: ClientConfig): SandboxClient {
   const conn = { apiKey: config.apiKey, baseUrl: config.baseUrl }
+  // Region-resolved endpoints (key prefix → cell), the single source the SDK
+  // uses internally — so our direct preview/network calls hit the same cell
+  // as the SDK-backed sandbox ops instead of defaulting to the primary.
+  const resolved = resolveConfig({
```
pavitrabhalla, line 261: MCP's own config layer normalizes empty/whitespace `SUPERSERVE_BASE_URL` to `undefined`, but the SDK's `resolveConfig` re-reads the raw env (`opts?.baseUrl ?? process.env.SUPERSERVE_BASE_URL`) where a set-but-empty value counts as a real override. Result: `networkLog`'s `new URL(path, "")` throws where it previously fell back to the default, and a usw key silently gets the default sandbox host (the empty override bypasses region derivation). Trimming/normalizing the env fallback inside `resolveConfig` would close it for all SDK consumers.

packages/sdk/src/config.ts
```diff
@@ -201,6 +209,9 @@ function deriveSandboxHost(baseUrl: string): string {
     if (host === "api-staging.superserve.ai") {
       return "staging-sandbox.superserve.ai"
     }
+    if (host === "api-usw.superserve.ai") {
```
pavitrabhalla: This hostname is now encoded twice in the file — here and in `KNOWN_REGIONS`, which already holds both `baseUrl` and `sandboxHost` (mirrored in the Python SDK's `_derive_sandbox_host`). Since the map has everything needed, this branch could be derived from it (match the override hostname against map entries), leaving only the staging mapping special-cased. As written, the next cell needs two edits per SDK, and a missed if-chain edit sends explicit-base-URL users of that cell to the default host's wildcard 404s.

apps/console/src/lib/sandbox-host.ts
```diff
@@ -0,0 +1,39 @@
...
+ * `NEXT_PUBLIC_SANDBOX_HOST` so staging/dev can point at `staging-sandbox…`.
+ * Non-default cells resolve to their own host and ignore that override — a
+ * `usw` sandbox is never reachable via the `use` proxy.
+ *
+ * Mirrors the SDK's region map (packages/sdk/src/config.ts). `use` stays on
+ * the legacy `sandbox.superserve.ai` host until its proxy also accepts the
+ * `use-sandbox` form; add it here once it does.
+ */
+
+const REGION_SANDBOX_HOSTS = new Map<string, string>([
```
pavitrabhalla, line 20: The non-default-cell hosts are hardcoded to prod with no env override (the doc comment says this is deliberate, and today nothing outside prod mints region-prefixed ids, so no live path hits it). Flagging as a future footgun: the day a staging usw cell exists, a staging console will send terminal WebSockets — including the access-token subprotocol — to the production usw proxy, with no env var able to redirect it. A per-region override map (defaulting to these values) would keep the deliberate prod behavior while leaving an escape hatch.

packages/mcp/src/lib/previewUrl.ts
```diff
@@ -30,35 +28,19 @@ export class PreviewUrlError extends Error {
...
-  } catch {
-    // Invalid URL — fall through to the safe default.
-  }
-  return DEFAULT_SANDBOX_HOST
-}
-
 /**
  * Build the public preview URL for `port` on `sandboxId`.
  *
+ * `sandboxHost` is the sandbox's data-plane host (region-resolved by the SDK's
```
pavitrabhalla, line 37: Two related notes: (1) `buildPreviewUrl` is now a near-duplicate of the SDK's exported `previewUrl`, and the two already disagree — this accepts ports 1–1023, the SDK rejects them because the edge proxy refuses privileged ports, so MCP can emit URLs that never route. Importing the SDK's builder would collapse both. (2) The host comes from the key's region (fixed at client construction) while the console derives it from the sandbox id — same-team flows are consistent by construction, but a pasted foreign-cell id silently yields a wrong-cell URL here where every other tool fails loudly. Deriving from the id (like the console) would align the two.

pavitrabhalla (commented):
Reviewed end to end — the approach is right and matches the endpoints we verified during the west-cell work. Two things I'd fix before merge, both one-liners: the SDK version gate (inline at client.ts:19) and the last line of the PR description, which cites an internal tracker ticket — per CLAUDE.md's Public repo section, PR descriptions are public; describe the motivation in words instead. (Same applies to the head branch name if it's cheap to re-push.) The remaining inline comments are follow-up material; details in each.

meAmitPatil:
Thanks for the thorough pass. Addressed in the latest push:

**Before-merge:**
- **SDK version gate** — bumped `@superserve/sdk` to `0.8.0` and `MIN_SDK_VERSION` in `mcp-publish.yml` together, with a note at the `resolveConfig` call site that it's the new floor. An MCP release can no longer pass the gate against a published SDK that lacks `resolveConfig`.
- **PR description** — reworded to describe motivation in words; dropped the tracker reference. (Left the head branch name as-is — renaming would orphan this PR thread; happy to re-push to a fresh branch if you'd rather.)

**Correctness/quality:**
- **Empty `SUPERSERVE_BASE_URL`** — `resolveConfig` now normalizes empty/whitespace to unset in both SDKs, so it no longer becomes a broken empty override or bypasses region derivation. Added tests.
- **`deriveSandboxHost` duplication** — now derives the cell mapping from `KNOWN_REGIONS` (both SDKs); staging stays the one special-case. A new cell needs only its map entry.
- **MCP privileged ports** — `buildPreviewUrl` min port is now 1024, matching the SDK; added a test (and fixed the id-safety test to use a valid port so it exercises the id check, not the port).
- **`SUPPORTED_SHARED_HOSTS`** — added a comment: `usw` is intentionally per-subdomain for now (verified working) until the usw proxy's shared-host path is confirmed; noted where to add it.

**Follow-ups (agree these are separate):**
- **Console per-region override** — no staging-usw cell mints region-prefixed ids today, so nothing hits it yet; worth an env-overridable per-region map when that changes, but out of scope here.
- **MCP host from key vs sandbox id** — kept key-based, since MCP wraps the SDK and this matches how the SDK's own `previewUrl`/data-plane resolve. Aligning MCP + SDK + console on one derivation is a good follow-up but a broader change than this PR.

Suites (SDK/Python/MCP/console) green, typecheck clean.

pavitrabhalla (approved):

PR #282, opened 2026-07-15
console: findTeamById prefers the cell a team names as home

pavitrabhalla:
## Summary
Platform impersonation of a migrated team fails with 401s: the impersonation cookie pins the team's region at start, and `findTeamById` returns the first cell match in configured order — for a migrated team that's the detached fallback row in the old cell, whose keys are all revoked. Part of the multi-region cells effort.

## Decisions
- A row found in the cell it names as `home_region` is authoritative; a row naming a different home is treated as a pointer and used only when the home row is missing (covers the pre-purge rollback window).
- Rows with no `home_region` (pre-migration schema) behave exactly as before.

## Testing
- 3 new tests: home-row preference over an earlier pointer row, pointer fallback when the home row is missing, null-home_region compatibility.
- Full team-directory suite: 24/24.
- Root cause verified live against a migrated team (impersonation 401 → resolves after the equivalent data-side fix).

### DIFF
--- a/apps/console/src/lib/api/team-directory.ts
+++ b/apps/console/src/lib/api/team-directory.ts
@@ -287,8 +287,14 @@ async function findTeamInRegion(
  * as home is authoritative; a row that names a different home is only a
  * pointer, used as a last resort if the home row is missing — including when
  * the home cell itself is degraded and could not be read.
+ *
+ * A pointer naming a cell that is not configured is dangling: its
+ * authoritative cell was never consulted, and the region it would hand
+ * callers routes impersonation and key minting into the detached copy. That
+ * fails closed — the row is ignored (logged) rather than served.
  */
 export async function findTeamById(teamId: string): Promise<TeamRecord | null> {
+  const regions = configuredRegions()
   const { items } = await acrossCells(async (region) => {
     const team = await findTeamInRegion(region, teamId)
     return team ? [team] : []
@@ -297,6 +303,12 @@ export async function findTeamById(teamId: string): Promise<TeamRecord | null> {
   let fallback: TeamRecord | null = null
   for (const { home_region, ...record } of items) {
     if (!home_region || home_region === record.region) return record
+    if (!regions.includes(home_region)) {
+      console.error(
+        `team directory: team ${record.id} row in cell ${record.region} names unconfigured home cell ${home_region}, ignoring detached row`,
+      )
+      continue
+    }
     fallback ??= record
   }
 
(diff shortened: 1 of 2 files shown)


apps/console/src/lib/api/team-directory.ts
```diff
@@ -272,19 +274,29 @@ async function findTeamInRegion(
...
-    if (team) return team
+    if (!team) continue
+    const { home_region, ...record } = team
+    if (!home_region || home_region === region) return record
+    fallback ??= record
   }
 
-  return null
+  return fallback
 }
```
awille-ai, line 316: **High — fail-open to the detached source when the declared home cell is not configured.** A pointer row whose `home_region` is absent from `configuredRegions()` is never checked against its authoritative cell, yet this

awille-ai (commented):
Reviewed specifically against the impersonation flow, multi-cell security boundaries, failure modes, and scaling. I found one high-severity fail-open edge and one medium scaling/test concern. I am leaving this as COMMENT rather than REQUEST_CHANGES because there are no critical findings, per the requested review policy. I am not approving while the high finding remains.

apps/console/src/lib/api/team-directory.ts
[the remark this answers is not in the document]
pavitrabhalla: Fixed in 2dc3c38. Rather than hand-rolling Promise.allSettled here, findTeamById now goes through the file's existing acrossCells helper, which already implements exactly this policy for the other directory lookups: parallel fan-out, secondary-cell failures degrade (logged), default cell still throws. Added tests for both failure paths, including home cell down → pointer row served.

pavitrabhalla: Fixed in 675e4cc: a pointer row whose home_region isn't in configuredRegions() is now skipped (logged) instead of served as the fallback, so the lookup returns null and callers fail closed — impersonation and key minting never route into the detached copy. A pointer only serves as fallback when its declared home cell is configured. Test added.

Your comment appears truncated — it ends mid-sentence at "yet this" — and the review body mentions a medium scaling/test concern I don't see posted. Can you repost both?

meAmitPatil (approved):

awille-ai (approved):
Re-reviewed at head 675e4cc. The previous high finding is addressed: dangling pointer rows whose home_region is not configured now fail closed instead of routing impersonation/key creation into a detached cell. The sequential lookup concern is also addressed by reusing acrossCells for parallel fan-out and secondary-cell failure isolation, with tests covering home-cell outage, default-cell failure, and unconfigured home regions. I do not see any remaining high or critical security, impersonation, scaling, or

PR #285, opened 2026-07-16
Fix admin access to impersonated audit logs

mohamedsobhi777:
## Summary

- add the dedicated `platform:activity:read` permission to console impersonation keys
- scope audit-log caches by the active or impersonated team
- resolve per-sandbox activity from the impersonated team's region
- fail closed with a clear authorization state when activity access is missing
- add permission, proxy-key, cache-isolation, and server-action tests

## Why

The audit-log page moved to the control-plane `GET /activity` endpoint, but temporary admin impersonation keys only carried sandbox and template scopes. The backend therefore rejected activity requests because impersonation keys intentionally have no customer RBAC actor. The per-sandbox activity action also fell back to the admin's own active team, and audit query keys were not isolated between team contexts.

## Impact

Authorized support admins can view another team's audit logs without receiving a 403 or mixing cached results between teams. Access remains read-only and requires the server-owned `platform:activity:read` authorization permission.

## Validation

- `bun run --cwd apps/console lint`
- `bun run --cwd apps/console typecheck`
- 58 focused Vitest tests passed across permissions, impersonation, proxy authentication, activity cache isolation, and activity server actions

### DIFF
--- a/apps/console/src/app/(dashboard)/audit-logs/page.tsx
+++ b/apps/console/src/app/(dashboard)/audit-logs/page.tsx
@@ -26,10 +26,13 @@ import { EmptyState } from "@/components/empty-state"
 import { ErrorState } from "@/components/error-state"
 import { PageHeader } from "@/components/page-header"
 import { Pagination } from "@/components/pagination"
+import { useQueryScope } from "@/components/query-provider"
 import { StickyHoverTableBody } from "@/components/sticky-hover-table"
 import { TableToolbar } from "@/components/table-toolbar"
 import { useActivityPage } from "@/hooks/use-activity"
 import { useListParams } from "@/hooks/use-list-params"
+import { useUser } from "@/hooks/use-user"
+import { canReadPlatformActivity } from "@/lib/admin/permissions"
 import { ACTIVITY_SORT_COLUMNS, type ActivityListParams } from "@/lib/api/types"
 
 const CATEGORY_TABS = [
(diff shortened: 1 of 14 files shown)


awille-ai (commented):
Reviewed together with sandbox PR #220 against the existing impersonation model.

I did not find a new cross-team access path or an authentication bypass.

Key points checked:
- `platform:activity:read` is added as a narrow, canonical impersonation scope; it does not imply sandbox, template, team-list, or write access.
- Starting impersonation still requires a Google-authenticated staff-domain user through `canStartPlatformImpersonation`; the permission alone is not sufficient.
- The main audit-log page continues through the existing proxied `GET /activity` path, where the backend enforces both the reserved impersonation-key identity and the activity scope.
- The existing per-sandbox server action remains a direct service-role database read, but this PR does not create that path. For impersonation it now authenticates with `getUser()`, resolves only a validated impersonation context, requires `platform:activity:read`, selects the impersonated cell, and filters by both `sandbox_id` and `team_id`.
- Audit React Query keys are now scoped by the active/impersonated team, preventing cached customer activity from appearing after a context switch.
- Missing activity permission fails closed in the impersonated UI and server action.

One non-blocking architectural note: the per-sandbox activity panel still uses a direct console service-role query while the full audit page uses the control-plane endpoint. That is pre-existing rather than newly introduced here, and the server-side team filter is correctly retained. Longer term, consolidating it onto the control-plane API would reduce duplicate authorization implementations.

No security-blocking finding from this review.

meAmitPatil (approved):

PR #287, opened 2026-07-17
Finish preview URL authentication surfaces

mohamedsobhi777:
## Summary

- Add explicit `public` and `private` preview URL policies to the TypeScript and Python SDKs, console, and MCP server.
- Preserve backend omission semantics: new publications inherit the sandbox default, while re-publishing an existing port without `access` preserves that port's mode.
- Replace console-local preview state with the backend publication API and mint a short-lived signed link only when the selected port is actually private.
- Keep public ports on clean credential-free URLs and preserve `legacy_public` compatibility only for older sandboxes.
- Reject the reserved boxd control port `49983` consistently across TypeScript, Python, MCP, console, and docs.
- Document publication, per-port access, token rotation, browser links, and machine header credentials across product docs and package READMEs.
- Protect signed preview credentials from iframe capability escalation and analytics/session-recording capture.
- Require the SDK release containing the new preview APIs before publishing the MCP package.

## Backend stack / dependency

This client PR deploys after the complete backend stack:

1. [sandbox#239](https://github.com/superserve-ai/sandbox/pull/239) — explicit preview-port publication
2. [sandbox#240](https://github.com/superserve-ai/sandbox/pull/240) — per-port public/private policy
3. [sandbox#242](https://github.com/superserve-ai/sandbox/pull/242) — per-port header-token authentication
4. [sandbox#243](https://github.com/superserve-ai/sandbox/pull/243) — browser signed-link authentication

The backend stack follows [the staged review on sandbox#201](https://github.com/superserve-ai/sandbox/pull/201#issuecomment-5007470378). No product feature flags are used; the PR/deploy boundaries provide the rollout order.

## Impact

New console and MCP sandboxes opt into strict published-port routing. Public previews expose only explicitly published ports. Private previews request a port-scoped credential and use the backend-advertised signed-link query carrier. Existing sandboxes continue to report the backend's `legacy_public` compatibility policy until a caller explicitly changes it.

The SDK packages should be released at `0.7.9` or newer before the updated MCP package is published.

## Validation

- Full monorepo `lint`, `typecheck`, `test`, `build`, and `format:check` passed
- Console preview/publication tests passed
- TypeScript SDK tests passed
- Python SDK tests passed
- MCP unit and surface tests passed
- Documentation navigation is synchronized with the local feature OpenAPI specification
- `git diff --check` passed


awille-ai (commented):
The backend and SDK now support a per-port access model (public/private) in addition to the sandbox default, but the console currently only exposes changing the sandbox default. Existing ports intentionally retain their own access mode, yet there doesn't appear to be a way to change an individual published port from public→private (or vice versa) without unpublishing and republishing it.

Is that an intentional product decision for Phase 1? If not, I'd consider exposing a per-port access toggle so the UI matches the underlying API capabilities. This isn't a blocker for me, but it feels like the product surface is incomplete relative to the backend.

awille-ai (changes requested):
I found one blocking behavioral issue in the Python SDK surface.

Both `Sandbox` and `AsyncSandbox` call `_require_live()` before every preview publication/token operation (`list_preview_ports`, `publish_preview_port`, `unpublish_preview_port`, `get_preview_token`, and `rotate_preview_token`). These endpoints are described and implemented elsewhere in this PR as control-plane operations that do not resume a paused sandbox. Requiring the instance to be "live" makes the Python SDK reject valid operations on paused sandboxes before it ever calls the backend, unlike the TypeScript SDK and MCP client.

That creates a cross-SDK inconsistency and blocks important workflows such as:

- changing publication state while a sandbox is paused;
- rotating/revoking a private preview credential without waking the sandbox;
- listing published ports for a paused sandbox.

Please remove the `_require_live()` guards from the control-plane preview methods (or replace them with a narrower guard that only rejects deleted/invalid handles, if one exists), and add sync + async tests covering at least listing and token rotation/publication on a paused sandbox instance.

Everything else I inspected is directionally solid: console state is backend-backed, private links use backend-selected query parameters, analytics/session recording receives explicit credential redaction, iframe capabilities are reduced, reserved port 49983 is consistently rejected, and CI is green.

awille-ai (approved):
Re-reviewed the latest head. The blocking Python SDK issue from my prior review is fixed.

The sync and async SDKs now use a narrower `_require_not_deleted()` guard rather than requiring an active VM for preview control-plane operations. That preserves deleted-handle safety while allowing listing, publication, token minting/rotation, and unpublication against paused sandboxes without resuming them.

The fix is covered in both sync and async tests, including explicit assertions that paused preview operations do not call `/activate` or `/resume`.

I also rechecked the broader change set:

- console preview state is backend-backed rather than localStorage-backed;
- per-port access mode controls whether a signed credential is minted;
- public links remain clean and credential-free;
- signed query parameters come from the backend contract rather than a duplicated constant;
- PostHog/session recording disables headers and bodies and redacts registered preview query parameters;
- embedded previews have reduced iframe capabilities and no referrer propagation;
- reserved control port 49983 is consistently rejected across console, TypeScript SDK, Python SDK, MCP, and docs;
- MCP publication/version ordering is documented and enforced by the release prerequisite.

CI and the PR-loop workflow are green, and there are no unresolved review threads.

Approved. Rollout should still respect the backend dependency order listed in the PR description, and the SDK release containing these APIs must precede the MCP package release.

PR #291, opened 2026-07-20
sdk: chunk stdin writes and surface exec close reasons

meAmitPatil:
When an exec WebSocket closed before a command finished, both SDKs raised a generic "connection closed" error that discarded the server's close code and reason. Separately, `stdin.write` sent arbitrarily large payloads as a single frame, which can exceed per-message limits on the server side.

- stdin writes are now split into 32 KiB frames transparently, in both the TypeScript and Python SDKs; ordering and byte-exactness are unchanged.
- Close errors now include the server's close code/reason. Code 1009 (message too big) maps to an actionable message suggesting the files API or smaller writes; an abrupt drop reports "connection dropped" instead of a bare code.
- One chunking test added per SDK.
- Versions bumped to 0.8.1 for release (also re-syncs the Python `__version__`, which had drifted from `pyproject.toml`).

### DIFF
--- a/packages/python-sdk/src/superserve/command_session.py
+++ b/packages/python-sdk/src/superserve/command_session.py
@@ -27,8 +27,10 @@ _CH_STDIN = 0
 _CH_STDOUT = 1
 _CH_STDERR = 2
 
-# Cap on a single stdin frame; larger writes are split transparently so no
-# frame ever approaches the data plane's per-message limit.
+# Cap on a single stdin frame; larger writes are split transparently. Must
+# stay under the smallest per-message limit enforced by any deployed server
+# (64 KiB historically; newer servers allow more) — raising it can break
+# sessions against servers that still enforce the older limit.
 _STDIN_CHUNK_BYTES = 32 * 1024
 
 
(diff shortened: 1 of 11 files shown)


packages/python-sdk/src/superserve/command_session.py
```diff
@@ -109,7 +113,9 @@ async def write(self, data: Union[str, bytes]) -> None:
         if not self._is_open():
             return
         raw = data.encode() if isinstance(data, str) else bytes(data)
-        await self._ws.send(bytes([_CH_STDIN]) + raw)
+        for off in range(0, len(raw), _STDIN_CHUNK_BYTES):
+            chunk = raw[off : off + _STDIN_CHUNK_BYTES]
```
pavitrabhalla: This chunks large payloads with an await self._ws.send(...) per chunk and no lock. Two concurrent write() calls (e.g. via asyncio.gather) with payloads over 32 KiB can have their chunks interleave on the wire, since the server just forwards each frame's payload straight to stdin with no reassembly. Doesn't that contradict "ordering and byte-exactness are unchanged" in the description? Is concurrent write() meant to be supported — if so, should this take a lock? (The TS SDK's equivalent loop has no await inside it, so it isn't affected.)

meAmitPatil: Right on both counts, fixed now.

pavitrabhalla (approved):
Nice fix — the Python lock around chunked sends is the right call since each `await ws.send()` yields, and skipping it on the TS side makes sense given that loop has no `await` points to interleave on. Also checked that `close_code`/`close_reason` are real attributes on the `websockets` library's `ClientConnection`, so the close-reason enrichment is grounded in the actual API.

One small thing: chunking via `range(0, len(raw), CHUNK)` means a zero-length `write("")`/`write(b"")` now sends nothing, whereas before it sent one frame with just the channel byte. Probably fine since nothing seems to depend on empty-frame semantics — was that intentional, or just an incidental side effect of the chunking loop?

meAmitPatil:
> Nice fix — the Python lock around chunked sends is the right call since each `await ws.send()` yields, and skipping it on the TS side makes sense given that loop has no `await` points to interleave on. Also checked that `close_code`/`close_reason` are real attributes on the `websockets` library's `ClientConnection`, so the close-reason enrichment is grounded in the actual API.
> 
> One small thing: chunking via `range(0, len(raw), CHUNK)` means a zero-length `write("")`/`write(b"")` now sends nothing, whereas before it sent one frame with just the channel byte. Probably fine since nothing seems to depend on empty-frame semantics — was that intentional, or just an incidental side effect of the chunking loop?

not a deliberate choice, but safe to keep

PR #294, opened 2026-07-23
console: read live sandbox counts from the sharded-counter view

meAmitPatil:
The control plane is moving the per-team active-sandbox counter from the `team.active_sandbox_count` column to a sharded counter table; after that migration the column freezes at its pre-migration value and live counts are served by the `team_active_sandbox_counts` view. This repoints the console's three readers — the customer quota banner and the admin team directory (list + single-team lookup) — at the view, defaulting to zero for teams with no counter rows yet.

All reads stay on the admin (service-role) client: the view is `security_invoker` over an RLS-deny table, so a user-scoped client would read empty rows rather than erroring.

**Merge ordering**: this must deploy after the control-plane migration creates the view in every cell's database — before that, the view queries fail. The old column keeps its stale value either way, so there is no window where the current code shows wrong data that this PR would have shown correctly.

### DIFF
--- a/apps/console/src/lib/api/quota-actions.ts
+++ b/apps/console/src/lib/api/quota-actions.ts
@@ -40,15 +40,23 @@ export async function getQuotaUsageAction(): Promise<QuotaUsageResponse | null> 
   if (!team) return null
 
   const admin = cellFor(team.region).createAdminClient()
-  const { data, error } = await admin
-    .from("team")
-    .select("active_sandbox_count, max_sandboxes")
-    .eq("id", team.teamId)
-    .single()
-  if (error) throw new Error(error.message)
+  // Live counts come from the sharded-counter view (team.active_sandbox_count
+  // is frozen); a team with no counter rows is absent from the view — zero.
+  // Admin client required: the view is security_invoker over an RLS-deny
+  // table, and a user-scoped client reads empty rows, not an error.
+  const [teamRes, countRes] = await Promise.all([
+    admin.from("team").select("max_sandboxes").eq("id", team.teamId).single(),
+    admin
+      .from("team_active_sandbox_counts")
+      .select("active_sandbox_count")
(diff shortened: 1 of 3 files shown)


meAmitPatil:
good catch — you're right on all three counts: the admin client bypasses RLS but not the API max-rows cap, the view's row population grows monotonically (shard rows persist at zero), and the query wasn't paginated. And the failure mode here is worse than the pre-existing unbounded team select: a truncated team list drops rows visibly, a truncated counts map shows a wrong 0 for teams that are present.

fixed by inverting the fetch: counts are now queried per id-chunk (`.in("team_id", …)`, 200 per request) for exactly the teams being returned — bounded by construction at any team count, and less data than the full-view read. single-team paths were already `.eq().maybeSingle()`. worth noting the team select itself still carries the older unbounded shape (pre-existing, sliced to 1000 post-fetch) — happy to paginate that too here or in a follow-up.

awille-ai (approved):
Rechecked the latest head. The earlier Supabase row-cap concern is addressed: `listAllTeams` now passes the concrete team IDs into `sandboxCounts`, which queries `team_active_sandbox_counts` in chunks of 200 using `.in("team_id", chunk)`. That avoids an unbounded view select and prevents silent truncation from being interpreted as zero active sandboxes.

The other readers remain correct:

- the customer quota action fetches the single team count with `maybeSingle()` and defaults a genuinely absent counter row to zero;
- the single-team admin lookup does the same;
- all count reads stay on the service-role client, which is required for the security-invoker view over the RLS-deny counter table;
- tests cover nonzero counts and the no-counter-row zero default.

The merge-ordering constraint still matters: deploy only after sandbox#225's migration has created `team_active_sandbox_counts` in every cell.

Lint, format, typecheck, docs-nav check, and tests are green on the current CI run; the build step is still in progress at review time. The code issue I previously flagged is resolved.

Approved.

PR #295, opened 2026-07-24
Aw/add admin billing

awille-ai:
## Testing

- [ ] Tested locally
- [ ] Added/updated tests
- [ ] All tests passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,93 @@
-# Platform Billing changes
+<div align="center">
+  <br/>
+  <br/>
+  <picture>
+    <img width="1500" height="auto" alt="Twitter Header - Personal" src="https://github.com/user-attachments/assets/ba41a4ed-1c8b-4826-bd3d-6fc8d84472ae" />
+  </picture>
 
-Copy these files over the matching paths in the superserve repository.
+  <br/>
+  <br/>
 
-Implemented:
+  <p><strong>Persistent and secure sandboxes for AI Agents, powered by Firecracker microVMs.</strong></p>
 
-- `/platform/billing` internal page
-- staff + `platform:billing:read` authorization
-- current-period totals across all customers
-- customer/team-name search
-- compact usage breakdown beneath each subtotal
-- credits applied, net due, and credits remaining
-- isolated per-customer API failures
-- focused action and component tests
+[![Docs](https://img.shields.io/badge/Docs-latest-blue)](https://docs.superserve.ai/?utm_source=github&utm_medium=readme)
+[![License](https://img.shields.io/badge/license-Apache%20License%202.0-blue)](https://github.com/superserve-ai/superserve/blob/main/LICENSE)
 
-Validation note: Bun and repository dependencies were not available in the execution environment, so lint, typecheck, and tests were not run here.
+</div>
+
+## Getting Started
+
+Visit [superserve.ai](https://www.superserve.ai/?utm_source=github&utm_medium=referral&utm_campaign=readme) or jump straight into the [docs](https://docs.superserve.ai/?utm_source=github&utm_medium=referral&utm_campaign=readme).
+
+## Structure
+
+```
+apps/
+  console/                 # Sandbox dashboard (Next.js 16, App Router)
+  ui-docs/                 # UI component docs (Vite)
+packages/
+  cli/                     # TypeScript CLI (@superserve/cli)
+  python-sdk/              # Python SDK (superserve on PyPI)
+  sdk/                     # TypeScript SDK (@superserve/sdk)
+  ui/                      # Shared UI components (@superserve/ui)
+  typescript-config/       # Shared tsconfig presets
+  tailwind-config/         # Shared Tailwind config
+docs/                      # Mintlify documentation site
+tests/                     # SDK end-to-end tests
+```
+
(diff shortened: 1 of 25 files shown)


awille-ai (commented):
I compared this PR against SS-245 and the merged sandbox API. I found a contract mismatch that should be addressed before merging.

The backend endpoint is `/internal/billing` (proxied by the console), accepts `search` rather than `q`, returns nested `summary` objects for successful rows, `error` objects for failures, `pagination` metadata, and `totals`. The UI currently calls `/platform/billing`, sends `q`, expects flattened fields (`region`, `billing_mode`, `compute_usd`, top-level totals/periods/total), and exposes `credits_applied_usd` as a sortable column even though the backend doesn't support sorting by it.

As written, this will not correctly interoperate with the merged backend.

I recommend:
- Use the actual backend route and query parameter names.
- Match the backend response shape exactly instead of maintaining a second frontend contract.
- Read summary cards from `totals` and pagination from `pagination.total`.
- Render row values from `summary` and `error`.
- Either add backend support for `credits_applied_usd` sorting or remove that server-side sort option until it exists.
- Add a contract test using an actual backend response payload so future API/UI drift is caught automatically.

On the positive side, this implementation does use a single paginated platform request rather than per-team requests, which satisfies the efficiency goal from SS-245 once the API contract is aligned.

awille-ai (commented):
I re-reviewed the latest head. The URL construction and most of the API contract mismatches have been improved, but the current proxy path still cannot authenticate to the sandbox internal API.

`/internal/billing` is behind the sandbox `InternalAuth` middleware. It requires `Authorization: Bearer <INTERNAL_API_TOKEN>` and the billing handler also needs `X-Actor-User-Id` to resolve and authorize the platform actor. The console proxy currently strips the incoming `Authorization` header, injects an `X-API-Key`, routes using the user's tenant cell, and does not set `X-Actor-User-Id`. That authentication model is appropriate for tenant endpoints, but not for an internal platform endpoint. It explains the runtime error `missing or invalid Authorization header`; this is not a missing RBAC permission.

Please give `internal/billing` a dedicated proxy/auth branch that:

- authenticates the console Supabase user and checks staff/platform billing permission;
- targets the configured platform/internal sandbox API base URL rather than a tenant/impersonated cell;
- sends `Authorization: Bearer ${INTERNAL_API_TOKEN}` from a server-only console environment variable;
- sends `X-Actor-User-Id: <Supabase user UUID>`;
- never forwards or exposes the internal token to the browser;
- bypasses tenant API-key lookup and impersonation behavior for this route.

Please also add proxy tests asserting the upstream URL and both internal headers, plus tests showing unauthenticated/non-platform users cannot use this proxy path.

awille-ai (commented):
Re-reviewed the latest head. The dedicated internal authentication branch is now aligned with the sandbox endpoint: it verifies the console user, injects `Authorization: Bearer <INTERNAL_API_TOKEN>` and `X-Actor-User-Id`, bypasses tenant API-key lookup, and targets `SANDBOX_API_URL`. That should resolve the previous `missing or invalid Authorization header` failure, assuming the console and sandbox deployments have the same `INTERNAL_API_TOKEN` configured.

There is still a blocking response-contract mismatch. The merged sandbox endpoint returns:

- `pagination: { limit, offset, total }`, not `{ page, page_size, total }`
- successful rows with `summary.cost_breakdown_usd.{compute,memory,storage}`
- `summary.billing_period.{start,end}`
- `summary.pricing_tier` and `summary.calculated_at`
- failed rows with `summary: null` and a populated `error`
- no `region` or `billing_mode` fields
- `totals` also includes `teams`, `succeeded`, and `failed`

The current TypeScript types and component still expect invented fields including `summary.region`, `summary.compute_usd`, `summary.billing_period_start`, `summary.billing_mode`, and `pagination.page_size`. A real successful response will display undefined breakdown/period values, and any failed row will crash when the component dereferences `row.summary.billing_mode` or `rowSummary.region` because `summary` is null.

Please model the wire response exactly, make `summary` nullable, render an unavailable row based on `row.error`/`summary === null`, use `cost_breakdown_usd` and `billing_period`, and derive the effective page size from `pagination.limit`. Update the tests to use a verbatim sandbox response fixture, including at least one `summary: null` error row; the current tests reproduce the frontend's invented shape and therefore cannot catch this incompatibility.

One acceptance-criteria gap also remains: SS-245 says Credits Applied must be sortable, but it is now intentionally rendered as a non-sortable header because the backend does not accept `credits_applied_usd`. That needs either a small follow-up to the sandbox endpoint or an explicit scope change to SS-245; client-side sorting only one page would not be correct.

awille-ai (commented):
Latest head still does not build. Vercel fails with `Type error: 'user' is possibly 'null'` at `headers.set("X-Actor-User-Id", user.id)` in `apps/console/src/app/api/[...path]/route.ts`. The earlier null check occurs in a separate control-flow branch, and TypeScript does not preserve the narrowing here.

Please retain the validated internal user in a non-null local variable, or guard immediately before dereferencing it. For example, inside the `if (internalBilling)` block:

```ts
if (!user) {
  return NextResponse.json(
    { error: { code: "unauthorized", message: "Not authenticated" } },
    { status: 401 },
  )
}
headers.set("Authorization", `Bearer ${internalToken}`)
headers.set("X-Actor-User-Id", user.id)
```

Once this builds, the branch alias can finally advance to the version that injects the internal bearer token. Until then it continues serving the older successful deployment, which explains why the UI still reports `missing or invalid Authorization header`.

PR #298, opened 2026-07-28
pr-loop: harden reviews without blocking CI

mohamedsobhi777:
## What changed

- Retry warm-sandbox connection failures up to three times with bounded 1s/2s backoff for server errors, request timeouts, and wrapped network failures.
- Generate separate pull-request and manual-dispatch steps so manual sweeps omit `--pr` entirely.
- Cap each workflow job at 30 minutes.
- Set `continue-on-error: true` on both loop review steps in the checked-in workflow, reusable template, and installer output.
- Document that review timeouts remain visible in logs without failing repository CI.
- Add regression coverage for retry behavior and generated workflow structure, including the advisory review setting.

## Why

PR reviews are useful advisory automation, but Claude rate limits and other external API failures should not turn unrelated repository CI red or block a critical change. Transient control-plane connection failures should also get a bounded retry before the review is abandoned.

## Impact

The Superserve repository's loop review becomes non-blocking when this change lands. After publishing a new `@superserve/loops` release to the `stable` channel, newly installed workflows will receive the same behavior. Review failures remain visible in the Actions logs for diagnosis and manual reruns.

## Validation

- `bun run test` — 49 tests passed
- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- Parsed both checked-in workflow YAML files and verified both review steps are advisory
- `git diff --check`


awille-ai (changes requested):
Re-reviewed the latest head at the same bar. The retry implementation, side-effect boundary, workflow generation, and tests are otherwise solid:

- retry is scoped only to warm `Sandbox.connect()`, before the review tick, so it cannot duplicate review side effects;
- retry is bounded to three attempts with 1s/2s backoff;
- final errors are preserved;
- permanent 4xx failures are not retried;
- server, timeout, and wrapped network failures are covered;
- pull-request and manual-dispatch workflow commands are correctly separated;
- manual dispatch omits `--pr`;
- generated workflows add a 30-minute job timeout and clarify the remote-secret-name contract;
- CI and the loop workflow are green, with no unresolved review threads.

One scope issue still blocks approval: `docs/docs.json` adds the preview-port API navigation entries (`GET/POST/DELETE /sandboxes/{sandbox_id}/preview-ports`). That change is unrelated to connection retry or workflow generation and belongs with the preview URL authentication/docs work. Keeping it here couples two independent changes, makes this PR's release scope less clear, and risks merge-order conflicts with the preview stack.

Please remove the unrelated `docs/docs.json` hunk from this PR (or explicitly retitle/rescope the PR if it is intentionally meant to carry that documentation change). Once that is removed, I do not see another blocker.

PR #300, opened 2026-07-28
mcp: bump to 0.1.3

meAmitPatil:
Bump the MCP server to 0.1.3 across all version surfaces: package.json, the SERVER_VERSION constant reported in the MCP initialize handshake and startup logs, and both fields of the registry manifest (server.json).

0.1.2 is skipped: that version is already on npm (published 2026-07-10) from a state where only package.json was bumped, so the installed server self-reports 0.1.1 while npm says 0.1.2. Publishing 0.1.3 with all surfaces aligned supersedes the mismatch; npm versions are immutable so 0.1.2 cannot be repaired in place.

### DIFF
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superserve/mcp",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "description": "Model Context Protocol (MCP) server for Superserve sandboxes — create, exec, and manage Firecracker microVMs from any MCP client",
   "keywords": [
     "agents",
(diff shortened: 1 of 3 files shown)


pavitrabhalla (approved):

awille-ai (changes requested):
The code change itself is correct and tightly scoped: all three version surfaces move together to 0.1.3 (`package.json`, `SERVER_VERSION`, and both registry-manifest fields), which fixes the 0.1.2 self-reporting mismatch described in the PR.

I do not see a code-level security or reliability issue in the diff. However, I cannot approve this head yet because it is currently reported as non-mergeable and the repository CI workflow is not attached to this commit. The only statuses present are the three successful Vercel preview deployments; there is no CI/loop workflow run for `f3a35a1939635e175da481fb543a777804fbbd9f`.

Please update/rebase the branch so GitHub can compute a merge result and ensure the normal CI workflow runs on the resulting head. Once the branch is mergeable and CI is green, this should be straightforward to approve.

PR #306, opened 2026-08-05
sdk: pre-warm the data-plane connection during create and connect

meAmitPatil:
`Sandbox.create()` and `Sandbox.connect()` now fire a fire-and-forget `GET /health` at the shared data-plane origin before their control-plane request, so the TCP/TLS handshake happens in parallel with sandbox creation/activation instead of in front of the first command or file operation.

- `sharedDataPlaneOrigin()` (config): the shared origin when shared-host routing applies, `undefined` in per-sandbox-subdomain mode (browsers, unsupported hosts). `dataPlaneTarget()` now delegates to it so the routing-mode logic exists once.
- `prewarmOrigin()` (http): fire-and-forget GET with a 5s abort; body drained so the keep-alive connection returns to the pool; all failures swallowed — if the ping fails, the first real request dials exactly as before.
- No-op in browsers and on self-hosted domains, where the per-sandbox origin isn't known until the sandbox exists.

Tests: new coverage for `sharedDataPlaneOrigin` and for the pre-warm call on create; existing assertions that indexed raw `fetch` mock calls now select the control-plane call explicitly.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superserve/sdk",
-  "version": "0.8.2",
+  "version": "0.8.3",
   "description": "TypeScript SDK for the Superserve sandbox API",
   "keywords": [
     "firecracker",
--- a/packages/sdk/src/Sandbox.ts
+++ b/packages/sdk/src/Sandbox.ts
@@ -15,10 +15,15 @@
  */
 
 import { Commands } from "./commands.js"
-import { previewUrl, type ResolvedConfig, resolveConfig } from "./config.js"
+import {
+  previewUrl,
+  type ResolvedConfig,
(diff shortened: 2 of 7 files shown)


meAmitPatil:
@codex review

awille-ai (approved):
Re-reviewed the latest SDK changes.

The pre-warm request remains intentionally best-effort: it is fire-and-forget, bounded by a timeout, drains the response to preserve keep-alive reuse, and swallows failures so it cannot change create/connect semantics. The shared-origin helper cleanly centralizes routing logic without changing browser or self-hosted behavior, and the updated tests correctly ignore the warm-up request when validating the control-plane calls.

I don't see any correctness, security, scalability, or API compatibility concerns. CI is green. Approved.

PR #311, opened 2026-08-20
docs: exempt console billing and stripe endpoints from the docs-nav check

meAmitPatil:
## Summary

The docs-nav check fetches the live OpenAPI spec, so when the new team billing and Stripe endpoints landed in the public spec, CI started failing on every branch in this repo. These are console-UI/webhook endpoints, not part of the developer sandbox API reference — the same category as the existing `/billing/pricing` exemptions — so this adds them to `EXEMPT_PATHS` rather than the public docs nav.

## Test plan

- `bun run docs:check-nav` now passes: "docs nav in sync with OpenAPI spec (57 operations)"

### DIFF
--- a/scripts/check-docs-nav.ts
+++ b/scripts/check-docs-nav.ts
@@ -26,6 +26,12 @@ const EXEMPT_PATHS = new Set([
   "/health",
   "/billing/pricing",
   "/billing/pricing/public",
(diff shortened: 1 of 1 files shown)


meAmitPatil:
Superseded — #304 added these endpoints to the docs nav, so the check passes on main again and the exemption isn't needed.

PR #310, opened 2026-08-20
sdk: add status, limit, and offset options to Sandbox.list

meAmitPatil:
## Summary

- `Sandbox.list()` in the TypeScript SDK and the Python sync/async SDKs now accepts `status`, `limit`, and `offset` alongside the existing `metadata` filter, passing them straight through to `GET /sandboxes`.
- All three options are additive and optional — omitting them keeps today's behavior exactly, so this is non-breaking.

## Why

The list endpoint has always supported status filtering and limit/offset pagination, but the SDKs only exposed the metadata filter, so every SDK call returned the team's entire sandbox list. Teams with a long sandbox history can now fetch just their running sandboxes (`status: "active"`) or page through the full list instead of downloading all of it on every call.

## Test plan

- TS: `npm run typecheck` clean, `vitest run` 234 passed (new cases: status/limit/offset in URL, no query string without filters)
- Python: `pytest` 309 passed, `mypy` clean (new cases in sync and async suites)

PR #312, opened 2026-08-21
sdk: bump TypeScript and Python SDKs to 0.8.3

meAmitPatil:
## Summary

Version bump ahead of releasing the new `Sandbox.list` options (`status`, `limit`, `offset`) from #310. Bumps all version carriers: TS `package.json` + `SDK_VERSION` in `http.ts`, Python `pyproject.toml` + `__version__` + `SDK_VERSION` in `_http.py`, and the regenerated `uv.lock`.

## Test plan

- Unit suites green on the underlying changes (234 TS / 310 Python)
- End-to-end against staging with both SDKs at this version: plain `list()` unchanged, `status` filter (string and Python enum) returns matching rows only, `limit`/`offset` pages are disjoint, combined filters round-trip

### DIFF
--- a/docs/sandbox/connect.mdx
+++ b/docs/sandbox/connect.mdx
@@ -28,8 +28,8 @@ sandbox.commands.run("ls /app")
 ## Find sandboxes with `list`
 
 List all sandboxes on the team, or narrow the result with filters. `metadata`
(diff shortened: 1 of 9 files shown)


meAmitPatil:
@codex review

awille-ai (approved):

PR #314, opened 2026-08-26
feat(console): add reCAPTCHA Enterprise check to signup

pavitrabhalla:
## Summary
- Adds a reCAPTCHA Enterprise check to the signup flow to cut down on automated account creation. The form generates a verification token on submit; the server verifies it via the Assessment API before creating the account.
- Adds the reCAPTCHA environment variables to the Turborepo build environment contract so Vercel builds receive the configured values.

## Technical decisions / tradeoffs
- Verification fails open on genuine transient integration/service availability problems, while verification failures that indicate an illegitimate request fail closed. Quota exhaustion and credential/configuration errors do not disable the CAPTCHA gate.
- The server action validates the token's shape at runtime rather than trusting the TypeScript signature, since server actions are directly callable with arbitrary payloads.
- Bounded timeouts on both the client-side token request and the server-side verification call, so neither can stall the signup form.
- When reCAPTCHA is configured but the client script cannot load or token execution fails, the signup is not submitted and the user gets an actionable bot-check load error.
- Staging and production use separately configured reCAPTCHA credentials/site keys. `RECAPTCHA_SCORE_THRESHOLD` controls the server-side score cutoff (default 0.5 when unset).

## Testing
- Typecheck, signup tests, lint, and format pass.
- Configured verifier coverage includes valid scores, invalid/action-mismatched/low-score tokens, quota/error handling, and timeout/failure policy.
- Manually exercised against the Vercel staging-target deployment:
  - fixed score 0 is rejected;
  - fixed score 1 proceeds through signup and creates the Supabase user;
  - simulated `grecaptcha.enterprise.execute()` failure displays the actionable bot-check error and does not submit signup.
- Verified the staging-target build artifact contains the configured public site key and reCAPTCHA Enterprise client code.
- Production reCAPTCHA environment variables have been configured in Vercel; production smoke testing remains to be performed after deployment.


apps/console/src/lib/recaptcha/verify.ts
```diff
@@ -0,0 +1,113 @@
...
+      )
+      // A malformed/forged token still gets a 200 with tokenProperties.valid
+      // === false (handled below) — Google validates token content inside a
+      // successful assessment, and the one HTTP-level bypass (an oversized
+      // token) is already rejected above by MAX_TOKEN_LENGTH. A non-OK
+      // response here means the *request* was rejected for a reason we
+      // don't control (bad API key, wrong project, site key/project
+      // mismatch, quota) — fail open so a config mistake degrades to "no
+      // CAPTCHA" rather than "no signups."
+      return { verified: true }
```
awille-ai, line 90: **Critical — quota exhaustion disables the abuse control.** `CreateAssessment` returns HTTP 429 / `RESOURCE_EXHAUSTED` when its quota is exceeded, but this branch treats every non-2xx as `{ verified: true }`. An attacker can repeatedly call the server action with ordinary-length junk tokens (which are sent for assessment), consume the assessment quota, and then get all subsequent signup attempts accepted once Google starts returning 429. That gives the attacker a direct way to turn off the CAPTCHA gate.

Please distinguish dependency/config failures here rather than blanket-failing open. At minimum, 429/quota exhaustion should fail closed; I'd also fail closed on credential/config errors such as 400/401/403 and reserve fail-open, if desired, for genuine transient availability failures (e.g. selected 5xx/timeouts). Google documents CreateAssessment quota exhaustion as 429: https://docs.cloud.google.com/recaptcha/quotas

apps/console/src/lib/recaptcha/verify.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (!Number.isFinite(parsed) || parsed < 0 || parsed > 1) {
+    return DEFAULT_SCORE_THRESHOLD
+  }
+  return parsed
+}
+
+// reCAPTCHA Enterprise. Fails open (treats the request as verified) whenever
+// the integration itself isn't configured or errors, so a misconfigured or
+// unreachable check never blocks real signups.
+export const verifyRecaptcha = async (
```
awille-ai, line 33: **High — the configured security boundary has no direct automated coverage.** This verifier is new, but no verifier test file is added. The existing `signUpWithEmail` tests don't set the reCAPTCHA env vars or mock `fetch`, so they only exercise the early `!apiKey || !projectId || !siteKey => verified` path. That means invalid tokens, action mismatch, low/missing scores, timeout/error handling, and especially the non-2xx fail-open behavior can regress without a test failing.

Please add focused unit tests for the configured verifier. The highest-signal cases are: valid score passes; invalid/action-mismatch/low score fail; and 429 does not turn verification into success. A timeout/failure-policy test would also make the intended availability tradeoff explicit.

awille-ai (changes requested):
Security/resilience, test-quality, and correctness review completed.

The core integration shape is good, but there is one security-control bypass that I think needs to be fixed before merge: quota exhaustion is currently converted into a successful CAPTCHA verification. Google documents that CreateAssessment returns HTTP 429 / RESOURCE_EXHAUSTED when quota is exceeded, so an attacker can burn assessment quota with normal-length junk tokens and then continue signing up once the verifier starts failing open.

There is also no direct test coverage for the configured verifier. The existing signup tests run with reCAPTCHA unconfigured, so they don't prove invalid/low-score/action-mismatch/HTTP-error behavior.

Google quota docs: https://docs.cloud.google.com/recaptcha/quotas
Google CreateAssessment docs: https://docs.cloud.google.com/recaptcha/docs/create-assessment-website

pavitrabhalla:
Both addressed in 392c08cd:

- Quota exhaustion (429) now fails closed instead of being folded into the general fail-open path — a caller burning quota with junk tokens no longer opens the gate for everyone else. Every other non-OK status (bad credentials, wrong project, etc.) still fails open, since those reflect our own setup rather than something a caller's request pattern can induce.
- Added direct test coverage for the configured verifier: invalid/expired token, action mismatch, score above/below threshold, missing score, threshold override and out-of-range override, 429 vs. other 4xx, and network failure — 15 new cases in `verify.test.ts`.

pavitrabhalla:
@codex review

awille-ai (commented):
Re-review of latest head `392c08c`:

- Previous Critical (429/quota exhaustion bypass): fixed; I resolved that thread.
- Previous High (no configured-verifier tests): fixed with `verify.test.ts`; I resolved that thread.
- Fresh security/resilience pass: no new abuse-control bypass found.
- Fresh test-quality pass: configured server verifier now has good focused coverage.
- Fresh correctness pass: one remaining High availability issue is already captured in the existing inline thread on `page.tsx`: when the configured client cannot load/execute reCAPTCHA, it sends `undefined`, and the configured server rejects that as `missing_token`. This means provider/script failure does not actually fail open as the PR description intends.

Once that client failure path is made consistent with the intended policy (and ideally covered by a page test), I don't see another High/Critical issue.

pavitrabhalla:
Addressed in faf6b88f:

- When a configured site key can't produce a token (script blocked, timed out, execute() failed), the form now shows an actionable message pointing at content/ad blockers instead of letting it fall through to the server's generic "couldn't verify you're human" rejection. Server-side behavior is unchanged — a missing token still hard-rejects, since that's also what a scripted signup skipping the browser widget looks like, and this feature's whole point is raising the cost of that.
- Also fixed a latent test-isolation gap I found while at it: the signup test suites relied on reCAPTCHA env vars being absent from the ambient environment rather than clearing them explicitly, so a dev/CI env with those vars set would've broken tests unrelated to what they're asserting. Verified by re-running the suite with the vars set — passes now, would not have before.

apps/console/src/lib/recaptcha/verify.ts
```diff
@@ -0,0 +1,115 @@
...
+      // it's something a caller can directly induce by sending enough
+      // requests, so failing open on it would turn "burn the quota with
+      // junk tokens" into a way to disable verification for everyone.
+      // Every other non-OK status reflects our own setup (bad API key,
+      // wrong project, site key mismatch) and isn't something a caller's
+      // request pattern controls, so those still fail open.
+      if (response.status === 429) {
+        return { verified: false, reason: "quota_exhausted" }
+      }
+      return { verified: true }
```
awille-ai, line 90: **High — credential/configuration failures still silently disable CAPTCHA.** The PR now states that quota exhaustion **and credential/configuration errors do not disable the CAPTCHA gate**, but this branch only fails closed for `429`; `400/401/403` and other non-OK responses return `{ verified: true }`. The new test even locks in `403 => verified: true`.

That means a revoked/bad API key, wrong project ID, site-key/project mismatch, or similar deployment/configuration mistake silently removes the signup abuse control for every request. Those are not transient availability failures and should fail closed (at minimum 400/401/403, plus 429). Reserve fail-open for genuinely transient dependency failures such as selected 5xx/network timeout cases.

awille-ai (commented):
Re-reviewed the current head with the usual bar. The earlier quota-exhaustion, verifier-coverage, and client-script failure issues are fixed, and exact-head CI is green. One High security-policy issue remains: configured credential/configuration failures still fail open, despite the PR's stated policy that these errors must not disable the CAPTCHA gate. See inline comment.

apps/console/src/lib/recaptcha/verify.ts
awille-ai: Fixed in 35f8506b: configured reCAPTCHA now fails closed for all 4xx responses, including credential/configuration errors and quota exhaustion. Transient 5xx, network, and timeout failures retain the documented fail-open policy. Added focused 403 and 503 coverage.

awille-ai:
@codex review

awille-ai (commented):
Re-reviewed current head. The prior credential/configuration HTTP handling issue is fixed: all 4xx responses now fail closed, while transient 5xx/network/timeout failures retain the documented fail-open behavior, with focused tests for 403 and 503. Exact-head CI is green.

One High remains in the current implementation, already called out in the open review thread: partial reCAPTCHA configuration still disables verification entirely because `!apiKey || !projectId || !siteKey` returns `{ verified: true }`. If, for example, the public site key is present but the server API key or project ID is missing, the browser behaves as though CAPTCHA is enabled while direct calls to the server action are accepted without verification. Fail open only when all three values are absent; any partial configuration should fail closed as a configuration error.

Holding approval on that remaining issue.

apps/console/src/lib/recaptcha/verify.ts
```diff
@@ -0,0 +1,118 @@
...
+  // a caller can send any JSON, not just what the TS signature promises.
+  token: unknown,
+  expectedAction: string,
+): Promise<{ verified: true } | { verified: false; reason: string }> => {
+  const apiKey = process.env.RECAPTCHA_API_KEY
+  const projectId = process.env.RECAPTCHA_PROJECT_ID
+  const siteKey = process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY
+
+  if (!apiKey || !projectId || !siteKey) {
+    return { verified: true }
```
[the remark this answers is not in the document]
awille-ai: Fixed in 519ec029: verification now fails open only when all three reCAPTCHA settings are absent. Any partial configuration returns a closed configuration_error result, preventing a missing credential from silently disabling the gate. Added coverage for the partial-config case.

awille-ai (approved):

PR #322, opened 2026-09-07
sdk(python): Retry self-clearing 409 on sandbox delete

saurabh850:
Follow-up to #308: mirrors the same retry logic in the Python SDK.

When a sandbox is deleted while mid-transition (e.g., resuming from pause), the control plane returns a `409 Conflict`. Like the TypeScript SDK, this PR introduces a `retry_conflict` flag to the internal HTTP layer that extends the backoff budget (to 11 attempts) specifically for this self-clearing state.

`Sandbox.kill` and `Sandbox.kill_by_id` (and their async variants) now pass `retry_conflict=True`, allowing them to transparently wait out the transition instead of failing fast and leaving orphaned sandboxes.

- **Sync & Async HTTP Layers:** Extended to support dynamic `max_attempts` growth on 409s.
- **API Surface:** Unchanged. `retry_conflict` remains an internal kwarg.
- **Tests Added:** Unit tests verify that a `409` correctly retries up to the conflict budget, while normal requests fail fast on terminal `409`s.

This brings the Python SDK into complete parity with the TS SDK's delete behavior.

### DIFF
--- a/packages/python-sdk/src/superserve/_http.py
+++ b/packages/python-sdk/src/superserve/_http.py
@@ -33,6 +33,7 @@ USER_AGENT = (
 
 # Retry tuning
 _MAX_ATTEMPTS = 3
+_CONFLICT_MAX_ATTEMPTS = 11
 _BASE_BACKOFF = 0.1
 _MAX_BACKOFF = 30.0
 _RETRY_STATUS_CODES = {429, 502, 503, 504}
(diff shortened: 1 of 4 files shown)


meAmitPatil:
Thanks for the contribution! I've cherry-picked your commit onto #308 so both SDKs ship the fix together. Closing this one now.

PR #333, opened 2026-09-10
sdk: pause() waits until the sandbox is paused, following the API's 202

meAmitPatil:
`pause()` sends `Prefer: respond-async` and returns as soon as the API accepts the pause; the sandbox finishes pausing on its own. `pause(wait=True)` / `pause({ wait: true })` instead follows it until `paused` under one deadline (`timeout` / `timeoutMs`) that covers every request, retry, backoff, and body read, with `signal` cancellation in TypeScript. `resume()` waits out a pause still in progress before resuming, so pause→resume keeps working.

Safe against today's API, which ignores the header and answers 204 as before; the 202 arrives with the control-plane pause-operation change.


awille-ai (commented):
High compatibility concern: this changes the public meaning of `await pause()` / `pause()` from “returns when the sandbox is paused” to “returns when the control plane accepts the pause.” Internal completion-dependent call sites have been updated to pass `wait: true` / `wait=True`, but external SDK consumers cannot be updated by this PR and can now observe `pausing`, race a resume/delete/reuse operation, or assume the sandbox has stopped when it has not. The control-plane API preserves synchronous behavior unless `Prefer: respond-async` is explicitly requested, so the SDK can preserve compatibility too: default `wait` to true and make the new asynchronous behavior opt-in (`wait: false`), or ship this as an explicitly versioned breaking change with migration guidance.

awille-ai (commented):
## Verdict

COMMENT — 1 High finding

I re-reviewed the current head with the full PR-review workflow. The earlier correctness/deadline/cancellation issues are addressed, and CI is green. One public-SDK compatibility issue remains.

### High — Preserve pause completion semantics by default

Both public SDKs still change the meaning of an existing successful call:

- TypeScript: `pause({ wait: false })` is the default.
- Python: `pause(wait=False)` is the default.

On the current/base SDK, `await sandbox.pause()` / `sandbox.pause()` returns only after the pause endpoint has completed. Once the control plane honors `Prefer: respond-async`, the same unchanged caller will return at `202 Accepted` while the sandbox is still `pausing`.

The PR has correctly updated the repository's own completion-dependent callers to opt into `wait: true`, but external customers cannot be migrated by this change. Existing code can reasonably do `await sandbox.pause();` and then assume the sandbox is paused (inspect state, recycle it, tear down surrounding resources, etc.). This PR silently invalidates that synchronization contract without a versioned breaking-change boundary.

I recommend preserving the released behavior by default (`wait: true`) and making acceptance semantics the explicit opt-in (`wait: false`). That gives customers the new non-blocking path without changing what existing source code means.

### Reviewer execution

| --- | --- | --- |
| Hostile/adversarial | sequential fallback | FINDINGS |
| Performance/resource efficiency | sequential fallback | PASS |
| Compatibility/rollout | sequential fallback | FINDINGS |

I also rechecked the previously reported Node 18 signal composition/listener cleanup, response-body deadlines, Python wall-clock timeout enforcement, auto-delete completion, per-poll timeout recovery, resume-during-pause handling, and abort propagation; those issues are fixed on the current head.

awille-ai (approved):
Follow-up after reviewing the linked product thread: the fire-and-forget default is intentional. The team explicitly agreed that customer-facing pause should return 202 immediately, complete in the background, and let resume handle a mid-pause sandbox. Given that clarified contract, my prior compatibility concern about `pause()` no longer waiting by default is withdrawn. I rechecked the rest of the current head and do not see another merge-blocking issue.

PR #335, opened 2026-09-12
feat(console): route /api/qm/* to qm-api and add the QM tenant data layer

pavitrabhalla:
## Summary
Groundwork for hosted QM in the console: the API proxy learns a second upstream (`QM_API_URL`, the separate qm-api service) for `/api/qm/*`, and a typed client + React Query hooks for QM tenants are added for the upcoming `/qm` screens. No pages use these yet; behavior of every existing route is unchanged.

## Technical decisions
- `QM_API_URL` is read per request; when unset, `/api/qm/*` returns a 503 with the standard error envelope instead of falling through to the sandbox API. Auth, impersonation `team_id` override, the read-only impersonation guard, and the header allowlist are shared with existing routes; path segments are re-encoded and dot segments rejected so a crafted path cannot escape the `/v1/qm/` prefix.
- `ApiError` gains optional `fields` for qm-api's `{ error, fields }` validation bodies; the existing `{ error: { code, message } }` path is untouched.
- Credentials never enter React Query state: the model provider key rides only in the request body (held per call in a WeakMap keyed by the variables object), and admin sign-in links are minted through a plain hook rather than a mutation, and via POST so read-only impersonation cannot obtain them.
- The hand-written types are pinned to a generated contract. `apps/console/openapi/qm.openapi.yaml` is a committed `/v1/qm/*` slice of the sandbox repo's `api/openapi.yaml`; `src/lib/api/qm.generated.ts` is generated from that slice; `src/lib/api/qm.contract.ts` asserts at compile time that the hand-written types, the client signatures, and each operation's HTTP method match it, so a contract change fails `typecheck` rather than failing at runtime. `bun run qm:check-openapi` regenerates and fails on any diff in CI. Vendoring the slice keeps generation deterministic and independent of a sibling checkout or a deployed service; the upstream spec location is configurable (`--spec` / `QM_OPENAPI_SPEC`). CI runs with the allow-missing-spec switch until the qm-api spec change is on main, and that switch expires on its own: once the upstream spec is reachable, leaving it set is itself a check failure, so wiring CI to the real spec cannot be forgotten.
- Cache writes are scoped to the active query scope: `patchTenant`, the delete snapshot/rollback, and the post-mutation invalidations address that scope's exact keys, so a mutation in an impersonation session cannot patch or evict another scope's cached tenants. One deliberate exception, commented as such: qm-api answers slug availability across every team, so a create invalidates that key in every scope.
- Impersonation keys now carry `platform:qm:read`. qm-api keys impersonated tenant reads to QM's own scope, so without it every impersonated `/api/qm/*` read is refused.
- Polling stops on a read error it cannot get past (401, 404, 409) rather than on any error: `retry` and `refetchInterval` are separate decisions, and a tenant left cached as transitional would otherwise be re-requested every two seconds forever. A transient failure keeps polling, because polling is what recovers from it.
- The QM cache scope is the impersonation scope paired with the active team, following the billing hooks. `useQueryScope()` alone is the literal `"self"` for every team a user owns and a team switch keeps the same QueryClient, so scoping on it alone would still let one team's tenant land in another team's list.
- QM reads and writes hold while the active team is unsettled. The team switcher flips the directory optimistically, before the cookie the proxy authenticates with changes, so a request issued in that window would be keyed to one team and executed as another — for a create, provisioning a tenant for the team the user just left, and for an admin link, minting a sign-in credential for it. Team creation carries the same marker, since it also switches, and both run in one React Query mutation scope so the selection has a single order instead of two writers racing for the same cookie. Creation also reconciles the directory with the new active team before it settles, on both sides of its refetch, since a degraded regional read comes back successful but partial. An admin link that resolves after the scope has moved is discarded rather than handed to the new session.
- Each mutation captures that scope in `onMutate`, because React Query re-reads a pending mutation's options on every render: a team switch or an impersonation change mid-request would otherwise land the response in whichever scope happens to be active when it settles.
- A create refreshes the list on every outcome, not only on success. The documented 502 leaves the tenant committed, so a failed create can still change what the list should show.

## Testing
- `bun run lint`, `bun run format:check`, `bun run typecheck`, `bun run test`, `bun run build` all clean.
- Console suite: proxy +12 tests (traversal, 401/503 precedence), client +2, QM hooks 33 covering the polling lifecycle including the deleted-tenant 404 and recovery from a transient failure, optimistic delete/rollback, per-scope isolation for delete and retry, an impersonation change and a team change mid-request, reads and writes held during a pending team switch, an admin link discarded when its scope moves, field errors, concurrent creates, and cache-free credentials.
- Local Codex review iterated to a clean pass.


awille-ai (commented):
Reviewed in the context of the hosted-QM architecture/POC and against SS-527, SS-528, and the dependent SS-517 contract. The overall direction looks good: this is productization around an architecture we've already validated (QM on Cloud Run, Superserve sandboxes underneath), and the proxy/security handling is generally solid. I found two important contract issues and one smaller cache-scope mismatch:

1. **High — SS-528's OpenAPI drift requirement is not implemented.** SS-528 requires the console QM types to be generated from or checked against qm-api's OpenAPI in CI, with drift failing the build. This PR adds handwritten mirror types but no generation/checking step. That matters because qm-api is being implemented concurrently and the contract is still moving.

2. **High — admin-link HTTP method currently disagrees with SS-517.** This PR calls `/qm/tenants/:id/admin-link` with `POST`, specifically so the console's read-only impersonation guard blocks credential minting. I think that security semantic is correct. However, SS-517 currently specifies `GET /v1/qm/tenants/{id}/admin-link`. Unless SS-517/OpenAPI is changed in lockstep, the console and qm-api will disagree at runtime. My recommendation is to make `POST` the backend/OpenAPI contract, then let the OpenAPI drift check above keep both sides aligned.

3. **Lower severity — the PR description overstates query-scope isolation.** The description says optimistic updates are scoped to the active query scope, but shared helpers for delete/retry (`patchTenant`, snapshots/rollback, broad invalidation) operate by unscoped `qmKeys` prefixes and can touch all cached QM scopes. I don't currently see a concrete cross-tenant data leak because tenant IDs are team-owned and impersonated writes are blocked, but either scope those helpers as claimed or soften the invariant in the description/comments.

I would leave this as **COMMENT**, not request changes: no blocker to the architecture, but I would resolve the first two before relying on this layer. CI is green, and the proxy routing/path validation/header stripping/model-key handling/admin-link non-caching all look well considered.

awille-ai (commented):
Re-reviewed current head `2a8321e` against SS-527/SS-528, SS-517, the prior review, and the hosted-QM architecture/POC.

The previous admin-link and query-scope findings are resolved:

- `admin-link` is now consistently `POST` in SS-517, the qm-api OpenAPI contract, the vendored slice, the compile-time method assertion, and the client call.
- Tenant cache mutation/snapshot/invalidation is now genuinely scoped to the active query scope, with regression tests proving another scope with the same tenant id is untouched. The broad slug-availability invalidation is intentional and correct because slug uniqueness is global.

The generated-contract work is also much stronger now: the vendored OpenAPI slice, generated TS, `qm.contract.ts` shape/method assertions, and CI regeneration check address the handwritten-type drift problem internally.

**One remaining High / acceptance-criterion gap:** CI currently sets `QM_OPENAPI_ALLOW_MISSING_SPEC=1`, so `qm:check-openapi` succeeds when the actual sandbox/qm-api OpenAPI source is unavailable. In that mode it proves only that generated TS matches the vendored copy; a change to qm-api's real `/v1/qm/*` contract can drift from that copy without failing this repo's build. SS-528 says the types must be checked against qm-api's OpenAPI in CI and drift must fail the build. The workflow comment says this escape hatch is temporary until the qm-api spec lands on main; I think that is the right fix, but I would treat removing the allow-missing path / wiring CI to the upstream spec as part of completing SS-528 rather than leaving it as an untracked follow-up.

Aside from that staged cross-repo contract check, the fresh hostile/security/test/correctness/performance/compatibility/data-state/observability passes did not find another material issue. Current CI is green, and the proxy routing, credential handling, polling termination, impersonation scope, and cache isolation all look solid.

PR #337, opened 2026-09-13
docs: scope the lint and format commands to workspace packages

pavitrabhalla:
## Summary

CLAUDE.md told contributors to fix formatting with `bunx oxlint --fix && bunx oxfmt --write` from the repo root. Following that literally reformats ~65 files — all of `docs/`, `guides/`, `pyproject.toml`, `CLAUDE.md`, `.oxlintrc.json`, the PR template — because those live outside every workspace. CI runs `turbo run format:check`, which is per-package and never sees them, so the instruction produces a dirty tree and drift that nothing enforces. An agent following it hit exactly this and had to revert the lot.

## Technical decisions

- Point at `bun run format`, which Turborepo runs inside each package, so the scope matches what `format:check` enforces.
- Say plainly that `bun run lint` only reports — the per-package script is `bunx oxlint` with no `--fix`, so the old command's autofix behaviour had to be called out separately.
- Name the three TypeScript workspaces with no `format` script (`packages/tailwind-config`, `packages/typescript-config`, `tests/sdk-e2e-ts`); nothing formats them today and pretending otherwise is how the drift started.
- Add `ruff format` to the Python line, scoped to the two Python workspaces. The line already claimed ruff did "lint + format" while no documented command ran the formatter, and unscoped `ruff format` rewrites files under `guides/managed-agents/`.

Not done here: adding `format`/`format:check` scripts to the workspaces that lack them, which would make CI enforce what this documents. That is a real fix rather than a doc change, and it wants its own PR.

## Testing

Verified each claim against `turbo.json`, the root and per-package `package.json` scripts, `.husky/pre-commit`, the `lint-staged` globs, and the root `pyproject.toml` uv workspace members. Docs-only change.


awille-ai (approved):
## Verdict
APPROVE

The documentation change matches the repository's actual formatting/lint topology. The root `format` command delegates to `turbo run format`, package workspaces that opt into formatting pair `format` with `format:check`, and the three explicitly named TypeScript workspaces (`packages/tailwind-config`, `packages/typescript-config`, `tests/sdk-e2e-ts`) currently define neither formatting script. The Python commands are also appropriately scoped instead of running Ruff across unrelated `guides/` content.

I also checked the pre-commit behavior: root `lint-staged` does run oxfmt on staged Markdown/JSON/etc. regardless of workspace, while CI's `turbo run format:check` only reaches workspaces that define the task. Calling out that mismatch is accurate and useful.

## Review scope
- Verified the root `package.json` Turbo and lint-staged commands.
- Verified `turbo.json` task scope.
- Verified the three workspaces called out as lacking format/format:check scripts.
- Checked that the new instructions do not claim `bun run lint` autofixes.
- Checked Python formatter scope against the intended Python workspaces.
- Documentation hygiene / contributor-command safety reviewed.

## Reviewer execution
| --- | --- | --- |
| Hostile/adversarial | sequential fallback | PASS |
| Documentation hygiene | sequential fallback | PASS |
| Compatibility/rollout | sequential fallback | PASS |

### Validation caveat
This PR itself has no `CI` workflow run because `CLAUDE.md` is outside the current CI `paths` filter; only the Vercel checks ran. For a docs-only contributor-instruction change, direct inspection of the referenced scripts/config is sufficient here, so I do not consider that a blocker.

PR #339, opened 2026-09-13
ci: run checks on pull requests that target a branch other than main

pavitrabhalla:
## Summary

CI's `pull_request` trigger is filtered to `branches: [main]`, so a stacked PR — one based on its parent's branch rather than main — gets no lint, typecheck, test or build run at all. It shows green because nothing ran. The QM console work hit this: #336 is based on #335's branch and has had zero CI coverage since it opened; the only check on it was a local run on someone's laptop.

Dropping the filter makes the job run on every PR regardless of base. GitHub tests the merge of the PR into its own base, so a stacked run legitimately covers the parent's changes as well, and the `paths` filter is untouched.

## Technical decisions

- Removed the `branches` filter from `pull_request` only. The `push` trigger stays pinned to `main`, so this does not add post-merge runs.
- The alternative — enumerating branch prefixes — would need upkeep every time someone picks a new naming scheme, and silently fails the same way when they don't.
- Left a comment on the trigger so the filter doesn't get reintroduced as a cost saving.

## Testing

Parsed the workflow to confirm `pull_request` keeps its `paths` filter and loses only `branches`, and that `push` is unchanged. The real proof is the first stacked PR that opens after this merges.


awille-ai (approved):
## Verdict
APPROVE

Removing the `branches: [main]` filter from the `pull_request` trigger is the correct fix for stacked PRs. GitHub evaluates a pull request workflow against the PR's target branch, so a stacked PR whose base is its parent branch is excluded by the old filter. Keeping the existing `paths` filter while removing only the base-branch restriction makes CI run on the stacked merge result without adding post-merge runs; the `push` trigger remains restricted to `main`.

## Review scope
- Verified the current workflow changes only the `pull_request` base-branch filter.
- Verified `push.branches: [main]` is unchanged.
- Verified the existing paths filter remains intact.
- Security/resilience: the job is an ordinary `pull_request` workflow on a GitHub-hosted runner; it does not request write permissions or consume repository secrets, so allowing PRs targeting non-main branches does not introduce a new privileged execution path.
- Compatibility: ordinary PRs targeting main behave exactly as before; stacked PRs gain the missing CI run.
- Checked the current head's CI run: `CI` completed successfully.

## Reviewer execution
| --- | --- | --- |
| Hostile/adversarial | sequential fallback | PASS |
| Compatibility/rollout | sequential fallback | PASS |
| Infrastructure/CI safety | sequential fallback | PASS |

### Validation caveat
This PR's own successful run is still a main-targeting PR, so it proves the edited workflow parses and executes but cannot itself demonstrate the stacked-target case. The stacked behavior follows directly from GitHub's `pull_request.branches` semantics; the first non-main-base PR after merge will provide the end-to-end confirmation.

PR #342, opened 2026-09-14
sdk: bump TypeScript and Python SDKs to 0.9.0

meAmitPatil:
Version bump ahead of releasing the pause changes from #333. Bumps every version carrier: TS `package.json` + `SDK_VERSION` in `http.ts`, Python `pyproject.toml` + `__version__` + `SDK_VERSION` in `_http.py`, and the `uv.lock` / `bun.lock` entries. The Cursor guides now call `pause` with `wait`, which only exists from this version, so their minimum pins move with it.

Minor rather than patch: `pause()` now returns once the pause is accepted instead of once the sandbox is paused, and `wait: true` / `wait=True` restores the old behaviour. A caret range on 0.8.x keeps the old contract until the consumer moves.


packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@superserve/sdk",
-  "version": "0.8.3",
+  "version": "0.8.4",
```
awille-ai: **High — use a version boundary for the intentional `pause()` contract change.** Existing npm consumers on a normal range such as `^0.8.3` will automatically accept `0.8.4`, so they can silently move from “`await pause()` means completed” to fire-and-forget. Publishing this as `0.9.0` (which the branch name already suggests) keeps `^0.8.3` consumers on the old contract while allowing callers to opt into the new release. If silent adoption by all compatible `0.8.x` ranges is explicitly intended, document that as the rollout decision.

awille-ai (commented):
**Verdict: COMMENT — 1 High.**

Review scope: release/versioning change for the TypeScript + Python SDKs that will ship the intentional fire-and-forget `pause()` semantics from #333. I ran hostile, security/resilience, test-quality, correctness, compatibility/rollout, and documentation-hygiene passes sequentially; only compatibility/rollout found a material issue.

All version carriers themselves are synchronized and CI is green. The High is the chosen release number: `pause()` intentionally changes its default completion contract, but publishing that as `0.8.4` puts it inside existing npm ranges like `^0.8.3`, which resolve to later `0.8.x` releases. Customers can therefore receive the breaking behavior without changing their dependency range. A `0.9.0` release creates the expected npm caret boundary (`^0.8.3` excludes `0.9.0`), and notably the branch is already named `sdk-0.9.0`.

If automatically moving existing `0.8.x` consumers to fire-and-forget is explicitly desired, this can be accepted as a deliberate rollout choice; otherwise I recommend shipping #333 as `0.9.0` rather than `0.8.4`.

One sequencing caveat: #333 is still open. Do not publish whatever version this PR reserves before #333 lands, because npm/PyPI versions cannot be republished later with the missing pause changes.

packages/sdk/package.json
meAmitPatil: Agreed, back to 0.9.0 in d3dedf5 so `^0.8.x` ranges keep the old contract. #333 is already merged.

awille-ai (approved):
Re-reviewed current head d3dedf5. The previous High is resolved: this now releases as 0.9.0, which preserves the old pause contract for consumers on ^0.8.x while making the intentional fire-and-forget behavior an explicit opt-in version boundary. #333 is already merged, so release ordering is also correct. I rechecked all version carriers, User-Agent constants, lockfiles, and Cursor guide minimums; they are synchronized and scoped correctly. CI is green. No remaining Critical or High findings.

PR #343, opened 2026-09-14
docs: pause returns on acceptance; wait returns once paused

meAmitPatil:
Documents the SDK 0.9.0 pause contract: `pause()` returns once the pause is accepted, `wait: true` / `wait=True` returns once the sandbox is `paused`, `pausing` is listed as a transient state, and `resume()` is noted to wait out a pause in progress. Option table added to the SDK reference.


awille-ai (approved):
Consolidated review (manual pass + PR-review skill): APPROVE.

The documentation matches the shipped 0.9.0 SDK contract: `pause()` returns on acceptance by default, `wait` follows through to `paused` (with delete-on-pause treated as successful completion), `pausing` is documented as transient, timeout/poll units are split correctly between TS and Python, TypeScript cancellation is surfaced, and `resume()` documents its wait controls.

One cross-PR note: #344 changes the meaning of `pollIntervalMs` / `poll_interval_s` during the first two seconds by polling faster than the supplied interval. If #344 lands with that behavior, this reference should be adjusted there (or in a follow-up) because the current wording correctly describes 0.9.0 but would no longer describe 0.9.1 exactly.

Reviewer execution (sequential fallback): hostile PASS; security/resilience PASS; test quality PASS (docs-only); correctness PASS; documentation hygiene PASS; compatibility/rollout PASS. CI is green.

PR #344, opened 2026-09-14
sdk: a waited pause checks every 50 ms for its first two seconds

meAmitPatil:
With `wait`, the SDK learns the sandbox is paused by reading its status, so a pause that finished at 500 ms was reported at the next one-second check. Now the first check is immediate and checks run every 50 ms for the pause's first two seconds, then at the caller's poll interval as before. A typical pause is reported within one check of finishing; long pauses cost the same number of reads as today.

Bumps both SDKs to 0.9.1.


packages/sdk/src/Sandbox.ts
```diff
@@ -489,8 +494,18 @@ export class Sandbox {
...
   ): Promise<void> {
+    // Monotonic: a wall-clock step must not hold the fast cadence open.
+    const started = performance.now()
+    let first = true
     while (true) {
-      await sleep(pollMs, ctx.signal)
+      if (!first) {
+        const young = performance.now() - started < PAUSE_FAST_POLL_WINDOW_MS
+        await sleep(
+          young ? Math.min(PAUSE_FAST_POLL_MS, pollMs) : pollMs,
```
awille-ai: **High — preserve an explicitly requested poll interval.** `Math.min(PAUSE_FAST_POLL_MS, pollMs)` means a caller passing `pollIntervalMs: 1000` (or even 5000) is still polled every 50ms for the first two seconds. The Python sync/async paths mirror this. That changes the public meaning of the option and can add ~40 GETs per waited pause before the caller-selected cadence begins, which compounds quickly under concurrent pauses. Keep the fast ramp for the default interval only, or otherwise expose/document a separate fast-ramp policy so callers retain a way to bound request frequency.

awille-ai (commented):
Consolidated review (manual pass + PR-review skill): COMMENT — 1 High.

The fast-poll ramp is correct on monotonic timing and the ambiguous-timeout race is fixed, but the implementation currently overrides the public `pollIntervalMs` / `poll_interval_s` contract during the first two seconds. A caller that explicitly asks for a 1s (or 5s) polling interval still gets status GETs every 50ms in that window. That is both a compatibility issue and a material request-amplification risk: one waited pause can add ~40 GETs before the normal cadence begins, and concurrent waited pauses multiply that directly. It also makes #343's new SDK reference inaccurate for 0.9.1, because that option is documented as controlling how often status is checked.

I would preserve the fast ramp only when the caller did not explicitly supply a poll interval, or otherwise change the API/docs so the caller can still bound request frequency. Add a regression test proving that an explicit slower interval is respected (or explicitly documenting/testing the new override semantics if that is intentionally the contract).

Everything else I rechecked is clean: immediate-first-poll behavior, early `active` tolerance after an ambiguous POST timeout, transition/error handling, monotonic clocks, version carriers/lockfiles, and CI.

Reviewer execution (sequential fallback): hostile FINDINGS; security/resilience FINDINGS (request amplification); test quality FINDINGS (no caller-interval contract assertion); correctness FINDINGS; performance FINDINGS; compatibility/rollout FINDINGS; sandbox lifecycle performance PASS apart from the same deduplicated polling root cause.

packages/sdk/src/Sandbox.ts
meAmitPatil: Agreed. Now the ramp only applies when no interval was passed; an explicit `pollIntervalMs` / `poll_interval_s` is honored from the first check. Tests added for both, and the reference table describes the default cadence. (#342 is merged, so this ships as 0.9.1.)

meAmitPatil:
@codex review

awille-ai (approved):
Re-reviewed at current head with the same two-pass bar (manual first, then PR-review skill). The previous High is fixed: an explicit `pollIntervalMs` / `poll_interval_s` is now respected as the caller-specified cadence, while the 50 ms fast ramp applies only when the caller leaves the interval unset. The docs were updated accordingly, so the contract remains accurate after #343.

I also rechecked the ambiguous-timeout path, monotonic timing, default ramp fallback after two seconds, resume behavior, version carriers/lockfiles, and the new regression coverage in TS + sync/async Python. The bounded default amplification is intentional and limited to waited pauses; I don't see a remaining Critical or High issue. CI is green.

PR #345, opened 2026-09-17
fix(console): unstick email sign-in and allow signed-out analytics

awille-ai:
Successful email sign-in could leave users on the sign-in page until they refreshed. The sign-in page can prefetch the signed-out home-page redirect, and the subsequent client-side navigation can reuse it. Use a full-page replacement after password authentication and when an existing session is detected, with validation that keeps the return destination on the same origin.

Exclude `/ingest` and its subpaths from auth middleware so signed-out PostHog requests reach the existing rewrite instead of receiving a 307 to the sign-in page followed by a 405. Similarly named routes remain protected.

Validation: 39 focused sign-in and middleware tests pass; console typecheck, changed-file lint, formatting, and diff checks pass. Production investigation observed a successful password response followed by no dashboard request, and analytics requests changing from 405 to 200 once the session was established. The fix has not yet been verified end-to-end in a deployed browser.


awille-ai:
@codex review

meAmitPatil (approved):

